Issue #11, opened 2017-10-05
Feature suggestion: display approval in Ark Votes instead of %


Jarunik:
Could we include approval in Ark Votes in addition/instead of the Approval in % in the delegate monitor?

Issue #1, opened 2017-10-04
Explorer delegate monitor


numba1enigma:
On page. Explorer.ark.io/delegateMonitor

Can the approval label become a buttor you can push. Which will toggle the votes between percentages and actual votes

For example flip between a value of 1% or 1,292,864 simply by pushing the approval label

Zarevok:
Agreed, it is nice to be able to see the vote amount

PR #18, opened 2017-10-07
Show Total Votes for Delegate

krau612:
Fix for #11 and #1

### DIFF
--- a/src/app/pages/address/address.component.ts
+++ b/src/app/pages/address/address.component.ts
@@ -29,7 +29,7 @@ export class AddressComponent implements OnInit, OnDestroy {
   public showBalanceFooter: boolean = false;
   public openVoters: boolean = false;
   public votersNumber: number = 4;
-  public supply: number;
+  public supply: number = 0;
 
   private _currentAddress: string = '';
   private subscription: Subscription;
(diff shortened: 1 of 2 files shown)


src/app/pages/address/address.component.ts
```diff
@@ -29,9 +29,11 @@ export class AddressComponent implements OnInit, OnDestroy {
   public showBalanceFooter: boolean = false;
   public openVoters: boolean = false;
   public votersNumber: number = 4;
+  public supply: number;
```
luciorubeens: Set a default value: `= 0`..

luciorubeens:
Can you take a screenshot how it will look? I can't reproduce here..

krau612:
Here's how the Approval line now looks:

screenshot at 2017-10-28 15-04-32


I just noticed that it always displays in Ark, should I make it display in the selected currency?

luciorubeens:
Thanks! Current token (ARK or DARK), take a look: https://github.com/ArkEcosystem/ark-explorer/blob/master/src/app/components/header/header.component.ts#L7

krau612:
So it should use initCurrency.name instead of currencyName and currencyValue? I think it would make more sense to use the currency name and value that the user has selected because I think every other value changes with the selected currency, such as Forged, Fee, Block Reward, etc.

luciorubeens:
Hmm it makes sense, I think it will seem a bit weird to see the total votes in USD

luciorubeens:
but yeah, to follow the standard you can use the current currency

krau612:
Sounds good :+1:

luciorubeens:
nice! +5 👍🏻

PR #19, opened 2017-10-08
Feature/transaction smartbridge tooltip

krau612:
Show a tooltip containing the Smartbridge text/ Vendor Field when hovering over the "Transaction ID" field of a transaction.

Also, fixed a bug where every transaction on the page for a block (https://explorer.ark.io/block/BLOCK-NUM) had the tag icon next to the Transaction ID, meaning it's Vendor Field contained text. However, the tag was showing even if the Vendor Field was empty.


luciorubeens:
Nice! thanks +5

PR #22, opened 2017-10-10
Fix mobile menu closing randomly because the currency gets set every …

krau612:
…few seconds.

The mobile menu would close every few seconds because the currency would be set every few seconds when the tickers update. The mobile menu would close, thinking that the user had selected a currency, even if no input has been given. This has been fixed by only auto setting the currency if it differs from the current currency (which should only happen once after loading).


luciorubeens:
+5 👍🏻

PR #25, opened 2017-10-19
Currency formatter (rounding, commas, and trimming insignificant 0's)

### LINKED ISSUE #9: Too much rounding.
In the new interface, it only shows 9.00 Ark 
In arkcoin.net, it shows the total value 9.00299711 Ark 
If the new interface shows the exact value, it would be great!

You should at least see the unrounded values somehow.

krau612:
Created a pipe to format currencies based on whether the selected currency is crypto (8 decimal places) or fiat (2 decimal places). It also strips 0's from the end of a number (ex. 123.50000000 -> 123.5). By default it uses the selected currency, but it can be overridden with an optional argument (for values that are always in one specific currency like in the header).

Resolves #9

### DIFF
--- a/src/app/pages/address/address.component.html
+++ b/src/app/pages/address/address.component.html
@@ -84,12 +84,12 @@
             <div class="ark-address-summary voters-container" style="display: block;" *ngIf="!openVoters" #voters>
                 <div class="voters-button" (click)="showBlock()"></div>
                 <span class="voters-address" *ngFor="let item of addressItem.voters; let i = index;">
-                    <a href="#" (click)="goToAddress($event, item.address)" *ngIf="i < votersNumber">{{item.address}} ({{(item.balance / 100000000) | currencyFormat: 'ARK'}} ARK)</a>
+                    <a href="#" (click)="goToAddress($event, item.address)" *ngIf="i < votersNumber">{{item.address}} ({{(item.balance / 100000000)*currencyValue | currencyFormat: currencyName}} {{currencyName}})</a>
                 </span>
             </div>
             <div class="ark-address-summary voters-container open" style="display: block;" *ngIf="openVoters" #voters>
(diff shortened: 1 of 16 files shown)


src/app/pages/address/address.component.html
```diff
@@ -84,12 +84,12 @@
             <div class="ark-address-summary voters-container" style="display: block;" *ngIf="!openVoters" #voters>
                 <div class="voters-button" (click)="showBlock()"></div>
                 <span class="voters-address" *ngFor="let item of addressItem.voters; let i = index;">
-                    <a href="#" (click)="goToAddress($event, item.address)" *ngIf="i < votersNumber">{{item.address}} ({{(item.balance / 100000000) | number: '1.2-2'}} ARK)</a>
+                    <a href="#" (click)="goToAddress($event, item.address)" *ngIf="i < votersNumber">{{item.address}} ({{(item.balance / 100000000) | currencyFormat: 'ARK'}} ARK)</a>
```
luciorubeens: Need to use a variable to handle this, because we can't manually change it to DEVNET.. the same discussed in #18

luciorubeens (changes requested):

boldninja:
@krau612 how is this looking?

krau612:
@boldninja Hey sorry for taking so long to get back to this, school and trying to get a job have taken up all of my time.

I've gone ahead and made the changes, but I couldn't test them because of some issue with res.ticker not being defined. That issue doesn't seem to be limited to this branch though.

faustbrian:
@krau612 Could you resolve the conflicts? Will take a look then and merge or let you know what should be changed.

faustbrian:
Will close this for now as there are quite a few conflicts, if you still wish to add this feature freel free to send a new PR.

PR #31, opened 2017-11-18
ngif everything in header individually

ckhatri:
I noticed on livesite that everything was filled except china, this way if something goes wrong that specific piece of data won't be shown and be empty.


luciorubeens:
+5 👍

PR #32, opened 2017-11-18
gives correct change over active period selected

ckhatri:
Noticed the explorer always says change over 1h, made it so its the change over the period.

So change over 1 day, 1 week, 1 month, 3 months, and one year.


luciorubeens:
sounds good! +5 👍

Issue #33, opened 2017-11-21
Option to remove price chart via cookie


dafty:
The current price chart takes up a lot of the screen, especially on smaller resolutions (not mobile).

Myself and a few others are not really interested in following the price of Ark, and would rather have an option to turn it off completely. Could we hide it via a cookie like the dark mode toggle?

Jarunik:
Upvote. That would be a good improvment.

PR #37, opened 2017-12-05
Toggle Price Chart & save in local storage

Nasicus:
Hey guys

Thought I'd help a little and give it a try.

I opened the PR already, because I was wondering what you guys think.
Basically I did the same as in switching the theme. However I cannot finde a good (font) icon for "no-chart" - so I'm wondering if we just want to use once icon (as I did) or if you guys have any other ideas ?
Icon search: http://fontawesome.io/icons/

This is still work in progress.

Enabled:
image

Disabled:
image

#33


Nasicus:
What we could also do to keep the theme consistent is to remove both icons of the "Dark-light-theme" switch and just add a single icon, like this one:
http://fontawesome.io/icon/star-half-o/

Just so you get a feel of it, this is how it could look like:
image

PR #42, opened 2017-12-09
Fix voters block (todo in code) & make it look better

Nasicus:
I'ts me again ;)

So there was this todo in the `address.component.html` to mege the two voters block.

First this is how it looked currently (though tbh at my local machine, there's also the balance there):
image

Well in my first commit  (8f47b6f) I completed the todo and merged the two blocks into one.
However I also did the following:
- change the icon nex to the address to somethingbetter
- change the expand icon so there is also a "unexpand" icon (see screenshot)
- ordered accounts by balance (so it's more meaningful!)

This is how it looks after my first commit:

changedlook

changedlook_expanded

However as you can see it still looks not good (at least in my opinion).
So I decided to take the same table as is used for top accounts.
I extracted the table to an own component and changed it so that it's a little more generic and can be used from other places.
Now it looks like this:
new look

Note that if you click "Load more" only 50 more voters are loaded each time. Because loading all voters at once into the DOM was slow anyway. However if this functionality is really needed / requested I'd suggest to make another button ("Load all").

Let me know what you think!

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -9,11 +9,13 @@ import { BlockComponent } from './pages/block/block.component';
 import { ActivityGraphComponent } from './pages/activity-graph/activity-graph.component';
 import { DelegateMonitorComponent } from './pages/delegate-monitor/delegate-monitor.component';
 import { TopAccountsComponent } from './pages/top-accounts/top-accounts.component';
+import { VotersComponent } from './pages/address/voters/voters.component';
 
 const appRoutes: Routes = [
   { path: '', component: ExplorerComponent, pathMatch: 'full' },
   { path: 'blocks/:page', component: BlockListComponent },
   { path: 'address/:id', component: AddressComponent },
+  { path: 'address/:id/voters', component: VotersComponent },
   { path: 'tx/:id', component: TransactionComponent },
   { path: 'block/:id', component: BlockComponent },
   { path: 'activityGraph', component: ActivityGraphComponent },
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -36,6 +36,8 @@ import { ToggleBackgroundComponent } from './components/header/toggle-background
 import { ThemeService } from './shared/services/theme.service';
 import { LocalStorageService } from './shared/services/local-storage.service';
 import { TogglePriceChartComponent } from './components/header/toggle-price-chart/toggle-price-chart.component';
+import { AddressTableComponent } from './components/address-table/address-table.component';
+import { VotersComponent } from './pages/address/voters/voters.component';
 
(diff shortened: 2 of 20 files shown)


faustbrian:
@Nasicus Could you resolve the conflicts? Will take a look then and merge or let you know what should be changed.

Nasicus:
@faustbrian done

Nasicus:
... here we go again - give me some minutes :>

faustbrian:
We talked about this PR and think it would be better to move this to a separate page like `/address/some-ark-address/voters` because of how long the list can get and keep `/address/some-ark-address` the way it is at the moment.

Meaning you would register a new route like `{ path: 'address/:id/voters', component: AddressVotersComponent },` and provide a new page this way.

Do you think you could move this to a new page that only contains the voters list?

Nasicus:
Hey

Just to clarify, why do you think it's getting to big?

In the current explorer it behaves like this:
demoold

In my version it is like this (by default 3 voters, when you click on more 50 additional are loaded):
demo

Of course I could move it to a seperate route (as you suggested). I think I just don't understand what you mean with "too long". And if I move it, I think we should still do something with the old logic? Also: How would you reach the "seperate page" ? By clicking on the "plus" on the right side? Or by clicking on a link like "Show voters list"?

faustbrian:
The voter list on the address details page itself should be kept short and only contain the addresses like it does at the moment as it already is quite long. If someone wanted more info about the voters they could go to the voters route and get a more detailed overview about who the voters are.

Could probably add a button like "View Voter Details" to get to the detailed list.

Nasicus:
Oki doki will do (with a few little tweaks which I already did in my first commit^^).

Nasicus:
Done - and I must say I kinda like it.

I did a small change in the voters-table (as to how it was before): it now shows how many % of votes each voter "weighs" (see screenshot).

As promised I left the existing list more or less alone, I changed the icon and did some refactoring (so the "todo" is removed).

Delegates view (votes closed):


Delegates view (votes expanded):


Voters page:


EDIT:
Just added a second commit where I use a "mono spaced" font, because then the voters list looks MUCH cleaner. Just tell me if you want this - or if you want me to use another mono space font!

image

faustbrian:
@Nasicus Last change I would like to request, could you add a "Load All" button similar to the "More" button to the voters page?

Nasicus:
Sure - the hardest part was, as so often, stupid CSS :P
Tell me if there's something there to optimize, CSS is not my favorite subject :P

It looks now like this:
image

Mobile:
image

faustbrian:
Thanks, looks good!

boldninja:
@Nasicus please provide me with your ARK address for Github bounty if you don't want it public please contact me on Slack (same username as here).

Nasicus:
0dsa8djh("/

Issue #43, opened 2017-12-10
Make model typed (maybe use ark-ts?)


Nasicus:
While doing my PR's I noticed that even though we use typescript a lot of stuff is not yet typed.
Especially all Ark model classes (blocks, users, transactions, addresses, etc.).

I'd like to make them typed.

This would be no problem, however I saw that there is already an ark-ts project:
https://github.com/ArkEcosystem/ark-ts

This project does not only define types but also offers methods to access the API.

Why is this project not used in ark-explorer?

I also noticed that the "model" and the api is not identical to the ark-explorer.

For example:

To get transactions:
Ark-Explorer: `/getLastTransactions`
Ark-ts: `/transaction`

Block model - number of transactions:
Ark-Explorer: `block.transactionsCount`
Ark-ts: `block.numberOfTransactions`

So in summary I have two questions:
- Should I make the model typed in ark-explorer?
- What is the connection / difference to `ark-ts` (and the used API)?

Sorry for the long post :)

PR #44, opened 2017-12-16
Add types for explorer api calls

Nasicus:
Had "some" time today and thought I'd address this ;)

I added types for all "normal" API calls and used them in the components.

Even if you'd like to include Ark-ts I thought this will be a good base... since we can probably relatively easy switch out the classes if we decide it.
=> And obviously you can always rejct this elsewise ;)

Note that I did not add any types for the delegate-monitor and for the activity-graph.
As far as I can see the model is not the same for the delegate-monitor (web-sockets) calls.
Should we add another model (other typed classes) there?

Addresses: #43


faustbrian:
@Nasicus Could you resolve the conflicts? Will take a look then and merge or let you know what should be changed.

Nasicus:
@faustbrian done 😅

PR #45, opened 2017-12-17
Fix all tslint errors & warnings

Nasicus:
Ok, this is a big one - but dont be scared :P

I always ignored the errors until now, but they are really annoying, especially if you use webstorm like me, there all tslint erros are shown as real errors when you have the editor open.

Anyway: I think they should be fixed, even if it's a lot.

In the most places the changes are pretty small and straight forward. So here are the imporant ones:

- Changed all components with wrong component-selectors (e.g. `app-block`) to ark prefixes (`ark-block`), for this I also had to adjust the `tslint` and the `.angular-cli.jso` file
  - I think this is an important change, since before there was a mix of `ark-` and `app-` components, and `ark` is obviously the real prefix
- Changed the selector of the `AppComponent` from `body` to `ark-app` (which is kind of the de-facto standard imho). I first didn't understand why this "strange" selector even existed, until I noticed it was to set the dark-theme class to the body tag. However there is a cleaner way to achieve this (see `AppComponent`) => I also had to adjust the `index.html` file for this to work again
- Apparently in LESS you have to specify a fallback font if you use a non-standard one, so I did this and used `serif` everywhere (note that this is NO change for the UI, tell me what you think about this and if you want another one, or not at all)
- Deleted all empty LESS files (in my opinion it's cleaner if a file only exists, if actually needed, but again, it's your project, tell me if I  should revert it)

The rest should more or less be small staff.

Btw: if you don't like some of the "fixes", we should adjust the `tslint` file accordingly instead.

I hope I didn't break anything - at least everything apperas to be working :)

(Looking forward to the merge conflicts with (my) the other PRs ❤️  )

### DIFF
--- a/.angular-cli.json
+++ b/.angular-cli.json
@@ -17,7 +17,7 @@
         "test": "test.ts",
         "tsconfig": "tsconfig.app.json",
         "testTsconfig": "tsconfig.spec.json",
-        "prefix": "app",
+        "prefix": "ark",
         "styles": [
             "styles.less",
             "../node_modules/font-awesome/css/font-awesome.css"
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -10,7 +10,6 @@ import { ActivityGraphComponent } from './pages/activity-graph/activity-graph.co
 import { DelegateMonitorComponent } from './pages/delegate-monitor/delegate-monitor.component';
 import { TopAccountsComponent } from './pages/top-accounts/top-accounts.component';
 
-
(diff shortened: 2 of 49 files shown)


j-a-m-l:
The code seems correct, but please, @luciorubeens, review this when you've some minutes

faustbrian:
@Nasicus Could you resolve the conflicts? Will take a look and let you know what should be changed.

Nasicus:
@faustbrian done

faustbrian:
@Nasicus Thanks, good work!

PR #46, opened 2017-12-18
fix(general.pipe.ts):Changes variable declarations as there are never…

cezarylos:
… reassigned, cleans redundant return of variable


cezarylos:
Closed due to previous PR which has the same changes.

PR #55, opened 2018-01-12
Fix error in search in header component

Nasicus:
reason:
getDelegate returned "undefined"

- rename methods to understand faster what they do

Note: It's less important that I thought, you have an error on the console, but the search itself does still work. Still not nice though.


faustbrian:
@Nasicus Please resolve the conflicts, will merge after that is done.

Nasicus:
@faustbrian done

Issue #52, opened 2018-01-12
Fix theme for toggle price chart button in dark mode


Nasicus:
![image](https://user-images.githubusercontent.com/1086065/34875654-7e31e28e-f79e-11e7-91d1-007c19c26b7c.png)

Will fix this myself :)

PR #57, opened 2018-01-12
Make toggle switch button look better in dark mode

Nasicus:
I realized that the toggle switch button looks horrible when disabled in dark-mode - it's way too shiny!

## Before
image

## Solution 1
Use same "dark" colors as are used for the light mode.
I find it's too dark then (edit: apparently highliy depends on the screen, on my desktop it doesn't look that bad) => not implemented
darken

Code:
```
:host-context(.dark-theme) {
  .tgl-item {
    color: @dark-header-menu-color;
    .tgl {
      + .tgl-btn {
        background: darken(@dark-secondary-color, 10%);
        &:after {
          background: @dark-secondary-color;
        }
      }

      &:checked + .tgl-btn {
        background: @switch-btn-checked;
      }
    }
  }
}
```

## Solution 2
Same as solution 1, but use `lighten` instead of darker for the background color.
I find it looks not very good => not implemented
lighten

## Solution 3
Use the color of the header bar as the background and the color of the font as the toggle button.
I like this best => implemented
new

Fixes: #52

### DIFF
--- a/src/app/components/header/shared/toggle-switch.styles.less
+++ b/src/app/components/header/shared/toggle-switch.styles.less
@@ -44,6 +44,10 @@
       position: relative;
       cursor: pointer;
       user-select: none;
+      background: darken(@light-secondary-color, 10%);
+      border-radius: 2em;
+      padding: 2px;
+      transition: all .4s ease;
       &:after,
       &:before {
         position: relative;
@@ -55,6 +59,9 @@
 
       &:after {
         left: 0;
+        border-radius: 50%;
+        background: @light-secondary-color;
+        transition: all .2s ease;
       }
 
       &:before {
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -10,6 +10,37 @@
...
+php artisan migrate:fresh
+php artisan migrate --path=tests/migrations --database=explorer
+composer play
+php artisan storage:link
+yarn run watch
+
+valet link explorer-ark-io
+```
+
+*Protip* Make sure you setup `EXPLORER_DB_*` to a local database in `.env` before running `composer play`.
```
ItsANameToo: can you add that it will need to point to a core database specifically

ItsANameToo: ok other approach: can you explain what EXPLORER_DB_* is it technically needs a core db defined there and it works with dummy data ONLY because you run that additional migrate command. Make sure it's clear in the readme what those things mean so you don't forget it after 2 weeks of not working on it :bleh:

PR #58, opened 2018-01-13
Add pagination for transactions & refactor

Nasicus:
@boldninja suggested to add pagination to all transactions views - and that's what I did.

I decided to add "real" pagination instead of infinite-scroll ("Load More").
For me the main advantage of this is that the DOM doesn't get too big and too slow and also that you can easily jump back to an very early state by adjusting the pagesize within the URL in your browser.

The look is the same as in the "Show all blocks" view.

However behind the sceners I introduced a `PaginationComponent` which is used by both "Show all transactions" and the address page for "All", "Sent" and "Received" transactions.

The new URL's look like this:

```
// All transactions (own page)
http://localhost:4200/transactions/1

// All address transactions (directly on the address page)
http://localhost:4200/address/DB8LnnQqYvHpG4WkGJ9AJWBYEct7G3yRZg/transactions/all/1

// Sent address transactions (directly on the address page)
http://localhost:4200/address/DB8LnnQqYvHpG4WkGJ9AJWBYEct7G3yRZg/transactions/sent/1

// Received address transactions (directly on the address page)
http://localhost:4200/address/DB8LnnQqYvHpG4WkGJ9AJWBYEct7G3yRZg/transactions/received/1
```

A sample screenshot of the address page:
image

Note that I refactored all transaction / addres transaction related methods in `ExplorerService` - I could delete quite a lot of duplicated code :)

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -10,11 +10,14 @@ import { ActivityGraphComponent } from './pages/activity-graph/activity-graph.co
 import { DelegateMonitorComponent } from './pages/delegate-monitor/delegate-monitor.component';
 import { TopAccountsComponent } from './pages/top-accounts/top-accounts.component';
 import { VotersComponent } from './pages/address/voters/voters.component';
+import { TransactionListComponent } from './pages/transaction-list/transaction-list.component';
 
 const appRoutes: Routes = [
   { path: '', component: ExplorerComponent, pathMatch: 'full' },
   { path: 'blocks/:page', component: BlockListComponent },
+  { path: 'transactions/:page', component: TransactionListComponent },
   { path: 'address/:id', component: AddressComponent },
+  { path: 'address/:id/transactions/:type/:page', component: AddressComponent },
   { path: 'address/:id/voters', component: VotersComponent },
   { path: 'tx/:id', component: TransactionComponent },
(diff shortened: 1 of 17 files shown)


src/app/app-routing.module.ts
```diff
@@ -10,11 +10,16 @@ import { ActivityGraphComponent } from './pages/activity-graph/activity-graph.co
...
 import { TopAccountsComponent } from './pages/top-accounts/top-accounts.component';
 import { VotersComponent } from './pages/address/voters/voters.component';
+import { TransactionListComponent } from './pages/transaction-list/transaction-list.component';
 
 const appRoutes: Routes = [
   { path: '', component: ExplorerComponent, pathMatch: 'full' },
   { path: 'blocks/:page', component: BlockListComponent },
+  { path: 'transactions/:page', component: TransactionListComponent },
   { path: 'address/:id', component: AddressComponent },
+  { path: 'address/:id/transactions/:type/:page', component: AddressComponent },
```
faustbrian, line 20: Is there a reason this line exists 3 times?

faustbrian (approved):

src/app/app-routing.module.ts
Nasicus: Yes. Code-blindness ;) 
Btw: I think I force pushed 1 minute before you commented and fixed this :P

PR #61, opened 2018-01-17
Show errors if an item cannot be found

### LINKED ISSUE #60: Give feedback if pages do not exist
E.g. if you go to a page with an invalid transaction:
https://explorer.ark.io/tx/a6c78629a644691b8dafc8e02c1c66a4ba3b8b45e1d94e1791d8f85248602b4b

At the moment we have a js error in the console, which isn't very helpful.

Nasicus:
Implemented for:

- transactions
- blocks
- addresses

Also refactored the corresponding methods in `explorer-service`  a little.

I implemented it so that the actual error is also displayed on the page (within a collapsible div). So the users can easily send us the real error messages if required.

Collapsed:
error collapsed

Expanded:
error expanded

fixes: #60

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -39,6 +39,7 @@ import { TransactionTableComponent } from './components/transaction-table/transa
 import { TransactionListComponent } from './pages/transaction-list/transaction-list.component';
 import { PaginationComponent } from './components/pagination/pagination.component';
 import { ClipboardComponent } from './components/clipboard/clipboard.component';
+import { ErrorSectionComponent } from './components/error-section/error-section.component';
 
 export function HttpLoaderFactory(httpClient: HttpClient) {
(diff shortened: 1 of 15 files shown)


faustbrian:
I think it would be better to simply redirect back to home page or show a neutral 404 page or at least drop the StackTrace.

Nasicus:
I find it clearer for the user if we show what the error was.
With dropping the stack trace I can live :D (so basically deleting the stacktrace)

So would you be fine with it if I remove it?

faustbrian:
Yeah, if you remove the stacktrace I will merge it.

Nasicus:
Rebased & write error to the console instead of showing it to the user.
Or do you prefer to not handle / log the error at all?

faustbrian:
Fine with me for now to log it to the console.

Issue #54, opened 2018-01-12
Delegate monitor title hidden on mobile


Jarunik:
![hidden-title](https://user-images.githubusercontent.com/4367481/34880075-8cace99c-f7af-11e7-9851-c70aea6b41da.png)

The currency bar is over the title of the delegate monitor.

Nasicus:
I wonder if we should display less info on mobile, e.g. only `ARK - USD` of the currencies ?

PR #62, opened 2018-01-19
Fix issue with header covering information

ItsANameToo:
At the moment, the fixed header covers some information on the explorer screens on mobile devices (e.g. top of the front page chart, the delegate monitor title, your the address when looking at your wallet, etc.). For now this can be fixed by increasing the padding a little to account for the larger header, although I don't think this is the proper solution. We should probably look for a different way of showing the information that is currently available in the header, as it takes up a lot of screen space. Some examples of before and after the change can be seen in the attached pictures.

Any suggestions as what would be a good alternative to the current header? We could make the header only show a single row, which can then be expanded to show the rest of the information. What are your thoughts on this or are there already some suggestions for this? Also; this change should solve issue #54 

Delegate screen - before:
delegate-before

Delegate screen - after:
delegate-after

Home screen - before:
home-before

Home screen - after:
home-after

### DIFF
--- a/src/assets/styles/_grid.less
+++ b/src/assets/styles/_grid.less
@@ -45,6 +45,6 @@ section {
 
 @media (max-width: @small-mobile) {
   section {
-    padding: 150px 10px 30px;
+    padding: 220px 10px 30px;
   }
 }


boldninja:
@ItsANameToo can you provide me your ARK address for GitHub bounty or contact me on Slack if you don't want to publish it publicly (username on Slack : 'boldninja' ).

ItsANameToo:
@boldninja I'll send you a message on slack!

ItsANameToo:
@boldninja 0)jsda9

PR #65, opened 2018-01-21
Clipboard feedback animation

wenright:
Displays a check mark next to the 'Copy to clipboard' tooltip after copying, as suggested by #59 
demo

PR #67, opened 2018-01-22
[WIP] Show message when search has no results

Nasicus:
I find it not very user friendly, that no feedback is shown when you search for something which has no results.
We should either redirect the user to a page or show a message to the user.

I decided to try the second approach and included ngx/toastr  for this:
resolve

We could also use this toastr in other places. One where I'd personally find it nice is for the clipboard success message.
I know that @wenright just implemented something in #65 . However for me the checkmark is too subtle. What would you think about displaying a success toast instead?

If we use the toaster I'd need to implement another placement for it on mobile (probably on bottom there).

### DIFF
--- a/.angular-cli.json
+++ b/.angular-cli.json
@@ -20,7 +20,8 @@
         "prefix": "ark",
         "styles": [
             "styles.less",
-            "../node_modules/font-awesome/css/font-awesome.css"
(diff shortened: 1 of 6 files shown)


faustbrian:
Toast sounds good to make it more obvious what happened. Send another PR for the clipboard.

PR #70, opened 2018-01-23
Add travis: delete all useless / not working tests, fix tslint, add .travis.yml

Nasicus:
I decided to make a clean base for ark-explorer concerning tests & eslint (again ;) ).
I think it would be best if we would start to use travis for all PRs!

The travis config does the following:
- Runs lint
- Runs tests
- Makes a prod build (some errors are not visible in non-prod builds)

I tested this PR on travis (on my own repo) and it succeeds!
So basically all you should need to do is to enable travis for this project.

Please don't merge this PR back before that is done, once you enabled it I will force push so a travis build is triggered and we can see if it works.

- [x] Activate Travis for ark-explorer


faustbrian:
Travis is now enabled.

Nasicus:
I force pushed but it doesn't build.
Do you think it will only work for new PRs? Should I close this one and try again?

faustbrian:
Should work with a new PR. All the GitHub integrations usually get activated when a PR is opened.

PR #71, opened 2018-01-24
Add Travis: Delete all useless / not working tests, fix tslint, add .travis.yml

Nasicus:
I decided to make a clean base for ark-explorer concerning tests & eslint (again ;) ).
I think it would be best if we would start to use travis for all PRs!

The travis config does the following:
- Runs lint
- Runs tests
- Makes a prod build (some errors are not visible in non-prod builds)

I tested this PR on travis (on my own repo) and it succeeds!
So basically all you should need to do is to enable travis for this project.

Please don't merge this PR back before that is done, once you enabled it I will force push so a travis build is triggered and we can see if it works.

- [x] Activate Travis for ark-explorer


Nasicus:
@faustbrian when I created the PR it did a build. However now I pushed again (to see if failing also works) and it doesn't build anymore. Maybe the build-triggers are not yet correct?

faustbrian:
The first build was OK so revert your `SHOULD FAIL` commit and I will check later what is up with Travis, on the website it shows that everything is connected and the opening PR went through.

Nasicus:
I think you may have to change the branch / pr policies now, because now it is building, but I think it's not mandatory for the build to pass before merge.

Anyway changed it and updated readme.

PR #76, opened 2018-02-03
Toaster styling + clipboard event mgmt

carrotderek:
- Made toaster messaging consistent with current look-and-feel
- Using built-in success callback for clipboard to display toaster feedback

toaster-success
toaster-info


Nasicus:
@carrotderek looks much better than what I did!
Is the `info` not too dark for dark mode?

carrotderek:
@Nasicus good point! I made a change in #78

PR #78, opened 2018-02-03
Made toaster styling more agnostic to dark and light themes

carrotderek:


resources/views/components/charts/price.blade.php
```diff
@@ -0,0 +1,63 @@
...
+<div x-data="makeChart()" x-init="renderChart()" class="bg-white border-t-20 border-theme-secondary-100 dark:border-black dark:bg-theme-secondary-900">
+    <div class="flex flex-col content-container-full-width md:py-16 md:px-8">
+        <div class="flex flex-col w-full">
+            <div class="relative flex items-center justify-between w-full">
+                <h2>Price</h2>
+
+                <x-ark-dropdown dropdown-classes="left-0 w-32 mt-3" button-class="w-32 h-10 dropdown-button">
+                    @slot('button')
+                        <div class="flex items-center justify-end w-full space-x-2 font-semibold flex-inline text-theme-secondary-700">
+                            <span>1 Week</span>
```
ItsANameToo: use `lang`, applies to all other instances too

resources/views/components/charts/price.blade.php
```diff
@@ -0,0 +1,63 @@
...
+                            <span :class="{ 'rotate-180': open }" class="transition duration-150 ease-in-out">
+                                @svg('chevron-up', 'h-3 w-3')
+                            </span>
+                        </div>
+                    @endslot
+                    <div class="py-3">
+                        @foreach (['daily', 'weekly', 'monthly'] as $period)
+                            {{-- TODO: Can't access `setPeriod` within that <x-ark-dropdown></x-ark-dropdown> component, so borked for now till I figure it out --}}
```
ItsANameToo: it's due to nested alpine yes, so your `this` is no longer pointed towards the scope where the marketchart stuff lives. Look into events and see if that can be used instead, as alpine abandoned the idea of being able to access parent instances

resources/views/components/charts/price.blade.php
```diff
@@ -0,0 +1,63 @@
...
+                    <div class="flex flex-col">
+                        <span class="font-semibold text-theme-secondary-500">Avg Price</span>
+                        <span class="font-semibold" x-text="priceAvg + ` ${currency}`">0.01570092 BTC</span>
+                    </div>
+                </div>
+            </div>
+        </div>
+
+        <div class="flex w-full" style="height: 340px;">
+            <canvas id="priceChart"></canvas>
```
ItsANameToo: this needs to be made unique so we can run 2 charts on the same page

public/js/chart.js
```diff
@@ -0,0 +1,266 @@
...
+                                            formatConfig.maximumFractionDigits = 5;
+                                        } else {
+                                            formatConfig.maximumFractionDigits = 3;
+                                        }
+
+                                        // return this.currency_format(price, formatConfig);
+                                        return `${value}$`;
+                                        /*return `$ ${(
+                                            value / scaleCorrection
+                                        ).toFixed(2)}`;*/
```
ItsANameToo: fix these things too, or remove the commented code

public/js/chart.js
```diff
@@ -0,0 +1,266 @@
...
+                                gridLines: {
+                                    color: themeColours.gridLines,
+                                    display: true,
+                                    drawBorder: false,
+                                },
+                                ticks: {
+                                    padding: 15,
+                                    ...fontConfig,
+                                    callback: (value, index, values) => {
+                                        // TODO: Proper implementation
```
ItsANameToo, line 1285: define proper, what is it that you need here

public/js/chart.js
```diff
@@ -0,0 +1,266 @@
...
+                                        if (
+                                            this.period !== "daily" &&
+                                            index === values.length - 1
+                                        ) {
+                                            return "Today";
+                                        } else if (this.period === "weekly") {
+                                            const width = this.$el.clientWidth;
+
+                                            if (width > 1200) {
+                                                // TODO: DW returns a day of the week, where value would be "Friday" for example
```
ItsANameToo, line 1327: todo

public/js/chart.js
```diff
@@ -0,0 +1,266 @@
...
+                                    );
+
+                                    if (index < midnight) {
+                                        return `Yesterday at ${title}`;
+                                    }
+                                    return `Today at ${title}`;
+                                } else if (index === values.length - 1) {
+                                    return "Today";
+                                } else if (this.period === "weekly") {
+                                    // TODO: We now works with dates for weekly such as 19.03
```
ItsANameToo: todo

resources/views/components/charts/price.blade.php
alexbarnsley:
```
<x-ark-dropdown :init-alpine="false">
```

Stops alpine from initiating on the dropdown, so the parent of the component changes

ItsANameToo: @Highjhacker try the suggestion of @alexbarnsley , he actually added that before so we can handle the dropdown ourselves instead of having it initialize another alpine instance. So you can use your parent, but need to add the commands to open/close the dropdown to it but at least won't result in this interference issue

ItsANameToo:
image
this should highlight the value that you have selected instead of showing all of them as unselected, check styleguide for it

ItsANameToo:
image
the circle thingy only shows up when you hover over one of the dots with your cursor, not when you move above it (while the tooltip does show)

ItsANameToo:
image

ItsANameToo:
image
:(

resources/views/app/home.blade.php
```diff
@@ -8,7 +8,7 @@
     @section('content')
         <x-general.search.header />
 
-        <div class="content-container-full-width justify-center md:py-16 md:px-8">
+        <div class="content-container-full-width justify-center py-16 px-8">
```
ItsANameToo: what's the use of `content-container-full-width` instead of the regular container :susp:

public/js/chart.js
ItsANameToo: @Highjhacker when will this get addressed? if not now: add a card that details what needs to be done

PR #83, opened 2018-02-07
Fix transactions counters

willdn:
Fix sender/recipient transactions counters.


faustbrian:
This is an issue on the end of the modified node. Will be fixed later today.

alexbarnsley:
@faustbrian this effectively matches the design (without the backend integration). I'm not sure how this will work in relation to the `performSearch` method in the SearchModule component.

PR #89, opened 2018-02-07
Don't truncate vendorfield on desktop

vmunich:
Truncated vendorfields look silly on the desktop:

before

I know that hovering the transaction id shows the full vendorfield in a tooltip, but I thought displaying the full field was more aesthetically pleasing:

after

Could also only truncate vendorfields longer than around 40 chars.

### DIFF
--- a/src/components/tables/Transactions.vue
+++ b/src/components/tables/Transactions.vue
@@ -26,7 +26,7 @@
           <wallet-link :address="transaction.recipientId"></wallet-link>
         </td>
         <td class="p-4 text-right border-none hidden lg:table-cell">
-          {{ transaction.vendorField }}
+          {{ truncate(transaction.vendorField || '', 35) }}
         </td>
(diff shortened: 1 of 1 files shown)


vmunich:
Thinking about it, should I open an issue and close this PR?

faustbrian:
The reason why we truncate the vendor field is that there are transactions with really long vendor fields which result in a table row stretching out to be as high as 2 rows.

What we could do is to increase the threshold at which the truncation appears. The `truncate` method accepts a second argument for that https://github.com/ArkEcosystem/ark-explorer/blob/master/src/mixins/index.js#L54-L58.

Just would need to test at which length it happens that a row splits into 2.

vmunich:
35 characters seems to be the spot, 36 chars or longer and it will split into 2:
36 chars or longer

faustbrian:
Yeah, seems to be fine when I check on a local install.

PR #96, opened 2018-02-09
Improved dark theme

ItsANameToo:
Slight improvement to the current dark theme, as the placeholder text and borders were very difficult to see on my screen.

Before:
heading-before

After:
heading-after


faustbrian:
The designer didn't agree with those changes but I made the placeholder slightly brighter.

ItsANameToo:
Fair enough. Would it be better to make an issue instead of a pull request for design-related issues in the future?

faustbrian:
Yeah, the designer makes the decision on those.

PR #99, opened 2018-02-10
Extract some Component CSS

Nasicus:
@faustbrian 

- How big you you want me to make these PR's? I think if they're getting to big it will get hard to review / test - what do you think?
- Is the table on the "delegates" page on purpose different (mainly other paddings I think) than all other tables??? (didn't extract anything there yet)
- How granular do you want the extracted stuff to be? 

Hope didn't break any styles :>


faustbrian:
1. Might be a good idea to send PRs for the extraction of specific pages/component types to keep it small in the future.

2. Yes, those tables look different on purpose.

3. I wouldn't go overboard and create a dozen special cases, just extract the really common ones to dry up the code and make it easier to re-use them in the future.

Will take a look at the changes in a bit unless you are going to push more changes.

Nasicus:
Thanks for the answers! Completely agree with 1 & 3.

I'm done for the moment. You can review by commit, I seperated them on purpose.

PR #100, opened 2018-02-10
Extract list-rows (mobile table) & more page sections

Nasicus:
Wondering if I should also create classes for stuff like `page-section py-10` which is now done quite often.

However how to call them?
`page-section-py-10` - but if we call it like that is there a benefit in it? Another idea would be to call them  `page-section-py-big`  or something like that.


faustbrian:
I wouldn't extract the spacing classes in those cases as they are a page specific detail and could change any time.

PR #103, opened 2018-02-11
Remove "next" when no further transactions are available

ItsANameToo:
When looking at the list of transactions in a wallet, you were able to keep pressing the "next" button even though there were no further transactions (and you were shown a "loading.." text even though it will never load any transactions, which is confusion). This PR adds a "totalTransactions" property that is passed to the paginator to ensure that the next button will be hidden on the last transaction page.

Question regarding this PR: I'm using two functions to retrieve the amount of sent/received transactions, but those are also calculated in `/src/components/wallet/Details.vue` as part of the wallet screen. Is there a way in which we can pass these values along as a property of some sort? As that would remove the need to make the same call again on the transactions page.

### DIFF
--- a/src/pages/Wallet/Transactions.vue
+++ b/src/pages/Wallet/Transactions.vue
@@ -8,7 +8,7 @@
       <div class="sm:hidden">
         <table-transactions-mobile :transactions="transactions"></table-transactions-mobile>
       </div>
-      <paginator :start="+this.page"></paginator>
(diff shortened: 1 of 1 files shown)


faustbrian:
Passing them along as property wouldn't be a pretty solution as it would make this component dependent on another one. Will leave it like this for now and refactor when Core 2.0 arrives.

PR #104, opened 2018-02-11
Add routes on home & extract tab css component

Nasicus:
- also "optimized" the top scrolling, by doing it the "vue-way" - or was there any reason it was done differently?
- don't know if `active-tab` and `inactive-tab` etc. are the best names, because there are a lot of other tabs too...didn't know a better name though

### DIFF
--- a/src/pages/DelegateMonitor.vue
+++ b/src/pages/DelegateMonitor.vue
@@ -8,18 +8,12 @@
       <nav class="mx-5 sm:mx-10 mb-4 border-b flex items-end">
         <div
           @click="activeTab = 'active'"
-          :class="[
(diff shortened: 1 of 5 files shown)


faustbrian:
Would merge the CSS refactor and scrolling but don't really see a benefit from the extra routes.

Nasicus:
Well tbh the reason I did it came up. While testing something in the latest block tab and I was annoyed that I had to switch the tab again after each page reload :P
But I can see that there's no real benefit for end users (though I also don't see any harm).
Anwyway: Should I revert and make a new PR then?

faustbrian:
Yeah, just send the CSS and Scroll changes.

PR #106, opened 2018-02-11
Also show red/green color on transactions overview

ItsANameToo:
This enables the red/green colors in the transaction overview that you see when you've clicked on the "Show more" button in a wallet.

Is it also an idea to add the red / green colors to the tables on mobile? I find that it makes it much easier to see whether a transaction was sent / received that way, but I'm not sure if it will result in a negative UX as you don't have rows in the mobile view.


faustbrian:
Asked the designer and he said OK to the mobile text colors.

PR #109, opened 2018-02-12
Add sent / received colors to mobile tables

ItsANameToo:
Adds red / green colors on mobile devices


faustbrian:
If you want to you could extract this into a component as this is now in a few places so would be a good candidate for a component like `<transaction-amount :transaction="transaction"></transaction-amount>`.

ItsANameToo:
I'll take a look into that! :)

PR #114, opened 2018-02-13
Localization of components

ItsANameToo:
This PR enables localization for the components in the explorer. Since it includes a lot of files, I'll be making a second PR later that contains the pages, after I've taken a small break from looking at this 😛 

As far as I could see, everything still works, but it wouldn't hurt if another person could take a look

Edit: pages will follow tomorrow


app/Http/Controllers/HomeController.php
```diff
@@ -13,6 +14,15 @@ final class HomeController
...
     public function __invoke(): View
     {
+        $monitor = new Monitor();
+
+        dd(
+            $monitor->status('0236d5232cdbd1e7ab87fad10ebe689c4557bc9d0c408b6773be964c837231d5f0'),
+            $monitor->status('025341ecfcbb48f9ecac8b87d6e5ace9fb172cee9c521a036f20861f515077bfc3')
+        );
+
+        // dd($monitor->lastForged('025341ecfcbb48f9ecac8b87d6e5ace9fb172cee9c521a036f20861f515077bfc3'));
```
ItsANameToo: sus

app/Services/Monitor.php
```diff
@@ -0,0 +1,100 @@
...
+
+final class Monitor
+{
+    public function activeDelegates(): Collection
+    {
+        return Wallet::query()
+            ->whereNotNull('delegate')
+            ->where('attributes->delegate->resigned', false)
+            ->orderBy('vote_balance', 'desc')
+            ->limit(51)
```
ItsANameToo: don't we have a `delegateCount` for this :bleh:, same for the other methods

app/Services/Monitor.php
```diff
@@ -0,0 +1,100 @@
...
+            return [
+                $round => [
+                    'min' => $roundStart,
+                    'max' => $roundStart + 50,
+                ],
+            ];
+        });
+    }
+
+    public function status(string $publicKey): Collection
```
ItsANameToo: I'm not sure about this method 🤔

PR #120, opened 2018-02-17
Replace hard coded ark tokens with network dependent token

Nasicus:
No idea if a `mixin` is the correct call to access the token, appears to be an easy way for me, if there's a better way, just let me know brian.


faustbrian:
Mixin is fine for now, will properly refactor that myself later on.

PR #121, opened 2018-02-18
Fix scripts on windows

Nasicus:
On windows the `dev:mainnet` and `dev_devnet` didn't work.
Because setting environment variables in windows has to be done via `set WHATEVER=ddd` .
However when I changed it to that it didn't work anymore on my ubuntu :P

So I used the [recommended way](https://webpack.js.org/guides/environment-variables/) from webpack to pass in custom arguments.

However I didn't adjust the `prod` config - wanted to hear what you think of it first.

Btw I recommend reviewing with ignoring whitespaces on this one:
https://github.com/ArkEcosystem/ark-explorer/pull/121/files?w=1


Nasicus:
oh didn't know... always did it like that in npm and just assumed it's the same in yarn :) => changed

faustbrian:
If you update the prod.conf I will test again and merge it if all works fine.

Nasicus:
Phew was harder & more complex that I thought 😅 
Hope everythings right now.

I unified the two configs so the arguments for both configs are parse the same way.

package.json
```diff
@@ -9,8 +9,8 @@
     "build:mainnet": "node build/build.js --network mainnet",
     "build:devnet": "node build/build.js --network devnet",
```
faustbrian, line 10: This should probably also be updated to use `--env.network` as you removed minimist.

package.json
```diff
@@ -9,8 +9,8 @@
     "build:mainnet": "node build/build.js --network mainnet",
```
faustbrian, line 9: This should probably also be updated to use `--env.network` as you removed minimist.

package.json
Nasicus: We can't use `--env.network` directly here, because it's a webpack thing I think. I didn't remove `minimist` it's in a central file now: `argument-parser.js` .
Same for other comments.

faustbrian: Ah yeah, overlooked that on mobile. Will test the changes tomorrow morning when I am back.

PR #124, opened 2018-02-25
Add LiteBit to Mainnet known wallets

j-a-m-l:


boldninja:
Address is same as Bittrex. LiteBit is using Bittrex to buy at market price and sends back to user probably.

j-a-m-l:
Lol, then this isn't needed.

faustbrian:
Should we change it to Bittrex / LiteBit then?

j-a-m-l:
I wouldn't do it because the real owner of the wallet is Bittrex and adding LiteBit would be inaccurate.

PR #125, opened 2018-02-26
Added Dutch translations

ItsANameToo:
Dutch translation for the explorer.

I haven't looked into adding a way to switch languages yet. Will that be the done the same way as on ark.io (so with the flag \> showing a list of available languages on hover)?


faustbrian:
We haven't added any translation selection to the explorer yet as it wasn't a major point on the to do for release.

ItsANameToo:
I understand, but I was wondering whether you guys already had something in mind on how it should be added (for example in the same style as on ark.io)?

faustbrian:
I will be adding that soon.

PR #133, opened 2018-03-10
Make hardcoded Ark related "price / currency" stuff configurable

Nasicus:
There were various prices were hardcoded stuff from Ark was shown, this removes hopefully all of this places.
Also it shows only market data if there is any.

Summary:
- make it configurable if cmc data should be displayed or not
- don't hardcode "ARK" for currency names
- hide Market cap & prices if there it's not available
- don't initialize currency state with "Ark" - make it configurable instead

Not sure though if the ` config` member is the correct place for all this settings.


faustbrian:
Really don't like how this is done, will take a look at this myself soon.

Nasicus:
And what do you not like?

PR #134, opened 2018-03-12
sanitize-tooltip

ItsANameToo:
Sanitizes tooltip containing smartbridge data


alexbarnsley:
Nice catch! Working, tested a couple other similar scenarios! Nice work 👍

PR #145, opened 2018-05-07
Always show previous / next text on block button

ItsANameToo:
On the block page (e.g. http://localhost:8080/#/block/2349249940958423570) the buttons to go to next / previous only showed the arrow and showed text when hovering over it. Personally, I did not like the buttons jumping around when hovering, so I changed it to always show the next / previous text. 

Unsure whether this is wanted / the best solution. An alternative I thought of would be to make use of the existing tooltips and show the next / previous text when you hover over the button (as they are only shown on desktop anyway). That way we can keep the smaller and sleeker buttons, but don't have them jump around when trying to show more information. Let me know what you think! :)

### DIFF
--- a/src/components/block/Identity.vue
+++ b/src/components/block/Identity.vue
@@ -6,12 +6,12 @@
       <div class="text-xl text-white semibold">{{ block.id }}</div>
     </div>
     <div class="hidden sm:block">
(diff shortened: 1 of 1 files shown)


faustbrian:
Will have to wait for the designers opinion on this, he will be back in a few days.

PR #146, opened 2018-05-07
Fix special transactions that were incorrectly colored as incoming

ItsANameToo:
This adds an extra check on `transaction.type` when providing a color to a transaction. Originally, special transactions such as votes, signatures- and delegate registrations were given a green color, as the recipient corresponded to the current wallet. However, they are actually outgoing transactions (and are also shown on the 'sent' page) and should therefore be indicated in red.

### DIFF
--- a/src/components/tables/Transactions.vue
+++ b/src/components/tables/Transactions.vue
@@ -33,7 +33,7 @@
 
       <table-column show="amount" :label="$t('Amount (token)', {token: networkToken()})" header-class="right-header-end-cell md:pr-4" cell-class="right-end-cell md:pr-4">
         <template slot-scope="row">
-          <transaction-amount :transaction="row"></transaction-amount>
+          <transaction-amount :transaction="row" :type="row.type"></transaction-amount>
         </template>
       </table-column>
 
--- a/src/components/tables/TransactionsDetail.vue
+++ b/src/components/tables/TransactionsDetail.vue
@@ -27,7 +27,7 @@
 
       <table-column show="amount" :label="$t('Amount (token)', {token: networkToken()})" header-class="right-header-cell" cell-class="right-cell">
         <template slot-scope="row">
-          <transaction-amount :transaction="row"></transaction-amount>
+          <transaction-amount :transaction="row" :type="row.type"></transaction-amount>
         </template>
       </table-column>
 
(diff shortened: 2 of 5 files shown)


geopsllc:
This change breaks mobile pages - transactions are not shown at all. Also, votes are still counted as both incoming and outgoing in the transactions field.

ItsANameToo:
This change is live on [explorer.ark.io](https://explorer.ark.io) and works as intended. If you are having issues on your fork, it might be because it is 68 commits behind and some other things are interfering.

Do you have any further information on the error that you get when applying the change from this PR? I don't see how it would break mobile pages as it simply adds a check for special transaction types.

geopsllc:
I see that it is live now and works on the ark explorer. Not sure why but it doesn't work on my fork. Works fine on desktop just not on mobile. Transactions are not showing up at all on mobile. Removing the additional check from the 2 mobile vue files start showing them again but with the wrong colors. Even on desktop the counts for in and out are still wrong but i guess that's calculated somewhere else.

ItsANameToo:
Looking at your commit, you made some mistakes when adding the changes to your fork. You used `:type="row.type"` in all the files, while the `mobile/transactions.vue` and `mobile/transactionsDetail` need `:type="transaction.type"` instead. 

That would at least explain why it gives you issues on mobile

geopsllc:
You are correct - I am a moron :)

ItsANameToo:
Not at all, it's easy to overlook small things 😛 

Regarding the counts for in and out, do you mean that votes are shown for both 'sent' and 'received'? Because I see that that is also an issue for the ark explorer then.

geopsllc:
Yup. The counter is counting them as both. Not sure where that calculation happens though.

geopsllc:
Thanks. I will try but I'm not sure I can figure it out. No clue how to add a filter by tx type. JS is not my thing :)

ItsANameToo:
Change the line to `.then(response => response.data.transactions.filter(transaction => transaction.type != 3))` and it will filter out the votes. However, I saw that this only removes it from the "received" tab, but are still counted as "received"..

Will take a closer look where the count is based on.

geopsllc:
Thanks. Maybe you can fix that for ark and push a commit when you have the time :)

ItsANameToo:
[This function](https://github.com/ArkEcosystem/ark-explorer/blob/master/src/services/transaction.js#L99) fetches the count, but at that point you cannot filter the transactions as we do not have those available (and some addresses will have hundreds of transactions that you would have to go through to filter the count correctly). Not sure whether there is a simple way to correctly show the count with the current API

geopsllc:
OK, thanks anyway. You're awesome!

PR #151, opened 2018-05-08
fixes sorting by rank in tables/Wallets.vue

dated:


faustbrian:
Everything Rank column related should be left aligned.

dated:
i put the alignment to the left back in place!

PR #156, opened 2018-05-08
Fix next / previous block buttons no longer showing on small screens

ItsANameToo:
Because of my other PR (#145) being merged after the one that added the buttons on small screens (#148), it got overwritten and they no longer appeared (sorry @dated). I've re-added them, with some slight adjustments, so they show up again.

Since the "previous block" text takes up quite some space, I've changed the media query to `768px` and the `sm` uses to `md`, otherwise it did not properly line up on screens that were between `sm` and `md` in size.


dated:
thanks for checking @ItsANameToo! i feared as much but had not the time to check ;) yea, the sm/md change is definitely necessary with the button text always shown! thanks for taking care of this.

PR #177, opened 2018-05-12
Use container to show currencies on screens below xl size

ItsANameToo:
Due to the amount of currencies in the explorer, the list would be too long to display on desktop screens that did not correspond to `xl` in size (see below image). Therefore this PR changes the header logic to show the container on `md` screens too.

currency-overflow

Additionally I've removed some unused styles that I came across in the header files


ItsANameToo:
closed in favour of #182

PR #178, opened 2018-05-12
Added a tooltip when the search returns no results

ItsANameToo:
When you search something but it has no result, you didn't get any feedback. I've added a tooltip that shows when nothing was found and dismisses automatically after 1500ms

search-tooltip

### DIFF
--- a/src/components/header/Search.vue
+++ b/src/components/header/Search.vue
@@ -13,6 +13,7 @@
       class="search-input w-full flex-auto mr-2 py-4 pl-4 bg-transparent"
       :class="{ 'text-grey': nightMode }"
       v-model="query"
+      v-tooltip="{show: nothingFound, content: $t('Nothing matched your search'), trigger: 'manual', placement: 'bottom-start', classes: 'search-tip' }"
       @keyup.enter="search" />
 
     <div class="search-icon text-grey hover:text-blue p-3 md:p-4" @click="search">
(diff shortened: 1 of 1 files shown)


resources/css/_general.css
```diff
@@ -284,3 +284,14 @@ body {
         @apply border-r border-theme-secondary-300;
     }
 }
+
```
ItsANameToo, line 287: you didn't copy the firefox stuff

```
.documentation-content .table-wrapper {
    scrollbar-color: #e5f0f8 #bad6f0;
}
```

resources/views/components/transaction-table-filter.blade.php
```diff
@@ -15,12 +15,13 @@ class="flex items-center justify-center w-5 h-5 font-semibold transition duratio
             </div>
         @endslot
 
-        <div class="items-center justify-center block py-3">
+        {{--w-60 will offset the navbar but then the button won't take the full width, which sucks--}}
```
ItsANameToo: what's the use of this comment :susp:

resources/css/_general.css
```diff
@@ -285,6 +285,26 @@ body {
...
+    background: transparent;
+    margin-top: 3rem;
+}
+
+.dropdown-scrolling::-webkit-scrollbar-thumb {
+    @apply rounded-lg bg-theme-primary-200;
+}
+
+.dropdown-scrolling {
+    scrollbar-color: #e5f0f8 #bad6f0;
```
ItsANameToo: this should also be transparent for background (if possible), check the docs https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color

ItsANameToo:
updated with changes, apparently firefox ignores our scrollbar anyway 🤷

PR #184, opened 2018-05-14
slight improvements to block page ux

dated:
while fetching the next block, the div containing the block id would get removed from the dom, resulting in a slight change in the height of the parent container. while fetching the next block, the fields containing a currency value would briefly display 'NaN Ѧ'

### DIFF
--- a/src/components/block/Identity.vue
+++ b/src/components/block/Identity.vue
@@ -4,7 +4,7 @@
       <img class="mr-6" src="@/assets/images/icons/block.svg" />
       <div>
         <div class="text-grey mb-2">{{ $t("Block ID") }}</div>
-        <div class="text-xl text-white semibold">{{ block.id }}</div>
+        <div class="text-xl text-white semibold">{{ block.id ? block.id : "&nbsp;" }}</div>
       </div>
     </div>
     <div class="flex w-full md:block md:w-auto justify-between">
--- a/src/mixins/index.js
+++ b/src/mixins/index.js
@@ -107,11 +107,13 @@ const methods = {
   },
 
   readableCrypto(value, appendCurrency = true) {
-    value = (value /= Math.pow(10, 8)).toLocaleString(undefined, {
-      maximumFractionDigits: 8,
-    })
+    if (typeof value != 'undefined') {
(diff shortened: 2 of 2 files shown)


resources/views/components/general/address.blade.php
```diff
@@ -17,10 +17,13 @@ class="flex items-center space-x-3"
                 {{ $prefix }}
             @endif
 
-            <a href="{{ route('wallet', $address) }}" class="font-semibold link">
-                <x-truncate-middle :value="$address" />
-            </a>
-        </div>
+        <a href="{{ route('wallet', $address) }}" class="font-semibold link truncated-address">
```
ItsANameToo: `sm:hidden md:flex`

resources/views/components/general/address.blade.php
```diff
@@ -17,10 +17,13 @@ class="flex items-center space-x-3"
...
+            <x-truncate-middle :value="$address" />
+        </a>
+
+        <a href="{{ route('wallet', $address) }}" class="font-semibold link truncated-address md:hidden">
```
ItsANameToo: `hidden sm:flex md:hidden`

resources/views/components/general/address.blade.php
```diff
@@ -17,10 +17,13 @@ class="flex items-center space-x-3"
...
-                <x-truncate-middle :value="$address" />
-            </a>
-        </div>
+        <a href="{{ route('wallet', $address) }}" class="font-semibold link sm:hidden md:flex">
+            <x-truncate-middle :value="$address" />
+        </a>
+
+        <a href="{{ route('wallet', $address) }}" class="hidden font-semibold link sm:flex md:hidden">
+            {{ $address }}
+        </a>
```
ItsANameToo: why was the `</div>` removed?

resources/views/components/transactions/list-mobile.blade.php
```diff
@@ -10,7 +10,6 @@
                                 <x-truncate-middle :value="$transaction->id()" />
                             </x-slot>
                         </x-general.loading-state.text>
-
```
ItsANameToo: was this removed for a reason? otherwise please put it back :hide:

PR #186, opened 2018-05-17
Show the delegate username in the Voters page

air1one:
To improve clarity on the Voters page, show the delegate username in the title.

As we go to the Voters page from the Wallet page, I preferred to get the username value from the Delegate component (and passing it to the Voters page) than making another request.

Tell me what you think 👍

### DIFF
--- a/src/components/wallet/Delegate.vue
+++ b/src/components/wallet/Delegate.vue
@@ -59,7 +59,7 @@ export default {
     getDelegate(wallet) {
       DelegateService
         .find(wallet.publicKey)
-        .then(response => this.delegate = response)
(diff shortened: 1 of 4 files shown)


faustbrian:
Looks good, just 1 small change regarding readability. If there are multiple statements in a code block you should break them into multiple lines to improve readability.

**Bad**
```js
.then(response => { this.delegate = response; this.$emit('username', response.username); })
```

**Good**
```js
.then(response => {
  this.delegate = response;

  this.$emit('username', response.username);
})
```

air1one:
Yes @faustbrian it's better breaking this up in multiple lines, done in the last commit. Thanks for the review 👍

PR #206, opened 2018-05-27
Update test-utils dependency

ItsANameToo:
The `shallowMount` function was introduced by PR #189. But this function is only available since the latest beta release of vue-test-utils; on older versions it will throw an error as it doesn't exist yet. Therefore this PR to update that dependency accordingly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "vuex-router-sync": "^5.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


ItsANameToo:
@alexbarnsley conflicts

alexbarnsley:
Thanks @ItsANameToo, updated

Issue #195, opened 2018-05-25
Improve performance of activeDelegates 


faustbrian:
The `activeDelegates` method has a major impact on performance as it sends hundreds of requests.

https://github.com/ArkEcosystem/ark-explorer/blob/a13b32652e0f3a647122375114e5adc3b27e25db/src/services/delegate.js#L106

This should be looked at to reduce the load on servers that are receiving requests from an explorer.

PR #210, opened 2018-05-29
Display forged blocks instead of forged arks in delegate monitor list...

air1one:
... and avoid making 51 requests at every timer refresh.

I started looking at #195 and although this does not concern activeDelegates method directly, it could help **a lot** reducing server requests.

To me, displaying forged blocks instead of arks makes sense, and we can still see forged arks in the delegate details.
Now tell me what you think 😃

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -77,7 +77,6 @@ export default {
     this.updateSupply()
     this.updateHeight()
     this.updateDelegates()
-    this.updateForged()
   },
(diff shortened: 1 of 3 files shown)


faustbrian:
Agree, this should reduce the load and if needed we could just do `producedBlocks * 2` later on for a rough estimate of forged ARK, excluding forged fees.

PR #213, opened 2018-05-29
adds pluralization to 'from transactions'

dated:
through this, the last-block component displays the singularized/pluralized form of `transactions` based on the count.

image
image
image


resources/views/components/general/username.blade.php
```diff
@@ -0,0 +1,28 @@
+<div>
```
ItsANameToo, line 1: if it handles both usernames and addresses, then `username` isn't a better name than `address` :bleh: call it `recipient` or `wallet-link` or something else that's more generic

ItsANameToo:
use a table component as tables have changed, and add whatever you need to the viewmodel instead of accessing raw data

PR #221, opened 2018-05-29
Add box shadow on block navigation button

ItsANameToo:
Adds a box shadow to be in line with other blue buttons:

From this:
schermafbeelding 2018-05-29 om 20 56 11

To this:
schermafbeelding 2018-05-29 om 20 59 30

### DIFF
--- a/src/styles/style.css
+++ b/src/styles/style.css
@@ -330,7 +330,7 @@ thead th {
 
 .pager-button:enabled:hover,
 .block-pager-button:enabled:hover {
-  @apply .bg-blue .text-white
+  @apply .bg-blue .text-white .hover-button-shadow-definition
 }
 
 .block-pager-button {


resources/views/components/blocks/table-skeleton.blade.php
```diff
@@ -0,0 +1,52 @@
...
+                </th>
+                <th class="text-right">@lang('general.block.amount')</th>
+                <th class="hidden text-right xl:table-cell">@lang('general.block.fee')</th>
+            </tr>
+        </thead>
+        <tbody>
+            @for ($i = 0; $i < 15; $i++)
+                <tr>
+                    <td>
+                        <div class="w-full h-4 rounded-md bg-theme-secondary-300 animate-pulse"></div>
```
ItsANameToo, line 23: `w-full h-4 rounded-md bg-theme-secondary-300 animate-pulse` can be a class instead, as it's copy pasted everywhere now :bleh:

faustbrian: I'll replace those components with skeleton components in another PR.

alexbarnsley: image

FYI

Issue #194, opened 2018-05-22
Vote tx considered both incoming and outgoing


geops22:
A vote transaction is considered both incoming and outgoing with the table entry having both red color and green color assigned. Only the second one is displayed which ends up being green.

dated:
This behaviour has been changed with https://github.com/ArkEcosystem/ark-explorer/pull/146

PR #224, opened 2018-05-30
Received transactions : remove votes

air1one:
See issue #194 for previous work on votes being considered as sent and received.

Something was left : we could still see the votes in the received transactions (and with a red color meaning sent).
So here filtering the votes transactions in receivedByAddress function.


ItsANameToo:
An issue with this is that the count on the wallet page still considers votes to be both sent and received. Therefore this change will introduce inconsistencies between the count and the amount of received transactions that are actually shown. I'm not sure whether that's desirable (@faustbrian ?), as it might result in wallets that have 20 incoming transactions, while the page only shows 10 due to the other 10 being votes that were filtered.

The count is fetched by the `receivedByAddressCount` function in the same file, but with the current API we cannot fetch the count without votes, as we cannot exclude them in the query due to the API `OR`ing all parameters. I have looked into this recently when discussing PR #146 , where it basically comes down to waiting for v2 before we can correctly handle this.

air1one:
Yes @ItsANameToo gave me a headache trying to see how we could get the count right 😅 but yes as you say in #146 I think we just have to wait for API v2.

So yes maybe we can wait v2 to do it all right (transactions + transactions count).

faustbrian:
As @ItsANameToo said, it would be better to wait for v2 to have better filtering capabilities. Will close this until v2 hits devnet.

PR #226, opened 2018-05-30
adds missing transaction types to link-wallet component

dated:
as can be seen [here](https://explorer.ark.io/transaction/2d8ed494d852880eb674c279fad64468bfccadd79d2a63cc69a5ca80116933f5) for example, some transaction types were not considered in the link-wallet component, leaving the "recipient" field empty. the missing [transaction types](https://github.com/ArkEcosystem/AIPs/blob/master/AIPS/aip-11.md#payloads) were added.

these are mostly placeholders, transaction type 4 is the only new one for which there actually are transactions.

if someone who speaks dutch wants to confirm the durch translations that be great :)

### DIFF
--- a/src/i18n/nl.json
+++ b/src/i18n/nl.json
@@ -118,5 +118,5 @@
   "IPFS": "IPFS",
   "Timelock Transfer": "Timelock Transfer",
   "Multipayment": "Multipayment",
-  "Delegate Resignation": "Ontslag Afgevaardigde"
+  "Delegate Resignation": "Aftreden Afgevaardigde"
 }
--- a/src/i18n/nl.json
+++ b/src/i18n/nl.json
@@ -113,5 +113,10 @@
 
   "2nd Signature Registration": "Registratie 2e Handtekening",
   "Delegate Registration": "Registratie Afgevaardigde",
-  "Vote": "Stem"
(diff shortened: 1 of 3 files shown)


src/i18n/nl.json
```diff
@@ -113,5 +113,10 @@
 
   "2nd Signature Registration": "Registratie 2e Handtekening",
   "Delegate Registration": "Registratie Afgevaardigde",
-  "Vote": "Stem"
+  "Vote": "Stem",
+  "Multisignature Registration": "Registratie Multisignature",
+  "IPFS": "IPFS",
+  "Timelock Transfer": "Timelock Transfer",
+  "Multipayment": "Multipayment",
+  "Delegate Resignation": "Ontslag Afgevaardigde"
```
ItsANameToo: Would be better to make it "Aftreden Afgevaardigde"

ItsANameToo (commented):
Dutch looks good, just one change needed :)

SpaceGuildIntern (commented):
Blijft toch gek klinken in het NL, (still sounds funny in dutch ) but yes agreed that should be it

dated:
Thanks! I'll change it asap

PR #228, opened 2018-05-30
adds tooltip showing the total votes to wallet page

dated:
image


samharperpittam:
I take back my approval.

- Set the fee chart toggle to off
- Refresh the page
- Turn the fee chart back on
- No chart

PR #230, opened 2018-05-30
Added tooltip with confirmations to well confirmed

ItsANameToo:
Adds a tooltip with the amount of confirmations if it is listed as "well confirmed":

schermafbeelding 2018-05-30 om 19 52 46

Change is only added to large screens currently, as we won't be able to see the tooltips properly on mobile devices anyway.

### DIFF
--- a/src/components/tables/TransactionsDetail.vue
+++ b/src/components/tables/TransactionsDetail.vue
@@ -47,7 +47,9 @@
               <img class="icon flex-none" src="@/assets/images/icons/clock.svg" />
             </div>
             <div v-else>
-              {{ $t("Well Confirmed") }}
+              <div v-tooltip="row.confirmations + ' ' + $t('Confirmations')">
+                {{ $t("Well Confirmed") }}
+              </div>
             </div>
           </div>
         </template>


resources/views/components/transactions/recipient.blade.php
```diff
@@ -2,18 +2,20 @@
...
+            @php($username = $transaction->recipient()->username())
+
+            <x-general.address :address="$address" :username="$username" />
         @elseif ($transaction->isVote())
-            @php($address = $transaction->voted()->username())
-            <x-general.address :address="$address">
+            @php($address = $transaction->voted()->address())
+            @php($username = $transaction->voted()->username())
+
+            <x-general.address :address="$address" :username="$username">
```
faustbrian: This component shouldn't be named `address` if it takes both an address and username. It also seems to always use the results of `->address()` and `->username()` so passing the model to the component will keep the views cleaner.

app/ViewModels/Concerns/Transaction/InteractsWithWallets.php
```diff
@@ -4,38 +4,40 @@
...
-            "transaction:wallet:{$this->transaction->recipient_id}",
-            Carbon::now()->addHour(),
-            fn () => $this->transaction->recipient
-        );
+        $wallet = $this->transaction->recipient;
 
-        if (is_null($wallet)) {
-            return $this->sender();
+        if (! $wallet) {
+            return $this->recipient();
```
faustbrian: This ends in an endless recursive call.

alexbarnsley: oops :hide: yeh was meant to be sender

resources/views/components/transactions/recipient.blade.php
alexbarnsley: yeh that's a fair point about naming and passing models in

PR #233, opened 2018-05-30
disables public view button if public key does not exist on wallet page

dated:
wallet pages for wallets without a public key, e.g. [AVytUqMWmdEhEY5gtcjoAPu88p5CtVBHid](http://localhost:8080/#/wallets/AVytUqMWmdEhEY5gtcjoAPu88p5CtVBHid), still display the button to switch to the public view, i.e. the address. this button has no function without the private view button counterpart, and should be disabled.

hiding the button all together looks off imo:

image

instead the the button is disabled, the color is set to grey and the cursor changed to default:

image

### DIFF
--- a/src/components/wallet/Details.vue
+++ b/src/components/wallet/Details.vue
@@ -54,6 +54,7 @@
       <div class="flex-none px-8">
         <button
           @click="view = 'public'"
+          :disabled="!wallet.publicKey"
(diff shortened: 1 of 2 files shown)


faustbrian:
Merging this to debug an odd issue that started some commits ago.

PR #235, opened 2018-05-31
displays chart tooltip with 8 decimals for cryptocurrencies

dated:
image

the tooltips currently show the currency symbol and the respective ISO code - what do you think about showing only either one?

furthermore i removed the changes introduced with #159 and #160, i'll upon an issue for that.


faustbrian:
I would only show the token name instead of the symbol.

dated:
image

PR #245, opened 2018-05-31
Fixed lint issues

ItsANameToo:
Fixed the lint issues. The `return await` statements are apparently [redundant](https://github.com/eslint/eslint/issues/7537)


faustbrian:
Yeah, if a method is `async` or returns `Promise` you can omit the await.

PR #251, opened 2018-06-01
add smartbridge icon to block transactions

reConNico:
small addition to the pr from dated: https://github.com/ArkEcosystem/ark-explorer/pull/250

Add smartbridge icon to block transactions

explorer_smartbridge_icon


ItsANameToo:
This also adds the icon to the "Latest transactions" table on the home page, as that one makes use of `src/components/tables/Transactions.vue` too

dated:
Here the smartbridge column is visible. The initial motivation was to show the icon if that was not the case, but i'm fine with adding the icon if there is a value, independent of whether there is the respective column.

PR #252, opened 2018-06-01
Moved smartbridge label icon to the front

ItsANameToo:
I liked the addition of the smartbridge label by dated (#250), but the placement results in the labels swerving around a bit due to the difference in space a transaction ID takes up:

schermafbeelding 2018-06-01 om 15 54 45

Therefore a proposed change: move the label to the front to keep it more consistent:

schermafbeelding 2018-06-01 om 15 58 35

It also adds an empty div to fill up the space of the icon if there is no smartbridge text to keep the transactions in line:

schermafbeelding 2018-06-01 om 15 53 08

Open to discussion 😄


dated:
I thought about this as well, the problem i saw is that when there are no transactions with smartbridge then there is a larger margin on the left than on the right. Does this take care of that case?

ItsANameToo:
At the moment it doesn't check for that, but I think it should be easy enough to add. First thing that comes to mind is to simply go over the list of transactions that we have for a given page (as it's 25 max) and check if any contain a smartbridge message. If so; do the current magic with the label, if not: remove the empty div to ensure that we don't have the extra empty space :)

dated:
My thoughts exactly, if you have time you can take care of it, i'm on the road for the next few hours :)

ItsANameToo:
I can take a look later today, but if you have the time to fix it before that; just go ahead

ItsANameToo:
I don't think you need another prop for it. Should be possible to use your `show-smartbridge-icon` prop for this purpose too

dated:
yep, i stand corrected. got a solution, but then there's this problem:

image

ItsANameToo:
What's the problem :p

dated:
ID and string are not aligned :D

image

ItsANameToo:
Id and icon are :P

Personally I find that better looking than the swerving of the label when you put it after the transaction id

Or try increasing margin when one of the transactions contain a smartbridge text, to align it with the number again

resources/views/app/transaction.blade.php
```diff
@@ -19,6 +19,8 @@
     @section('content')
         <x-transaction.header :transaction="$transaction" />
 
+        <x-marketsquare-banner />
```
faustbrian: This should only be shown if a network uses marketsquare.

resources/views/components/marketsquare/banner.blade.php
```diff
@@ -0,0 +1,31 @@
...
+    <div class="flex flex-row justify-center w-full p-8 rounded-lg sm:justify-between h-30 lg:h-32 bg-theme-primary-600 dark:bg-theme-secondary-900">
+        <div class="flex flex-row items-center justify-between space-x-4">
+            <div class="flex items-center">
+                <div class="w-12 h-12 border-white circled-icon text-theme-secondary-400">
+                    @svg('app-marketsquare', 'w-4 h-4 text-white')
+                </div>
+            </div>
+
+            <div class="flex flex-col justify-between space-y-2 font-semibold md:ml-4">
+                <div class="text-sm leading-tight text-theme-primary-300 dark:text-theme-secondary-700">For more {{ $transaction->typeLabel() }} details</div>
```
ItsANameToo: lang file

resources/views/components/marketsquare/banner.blade.php
```diff
@@ -0,0 +1,31 @@
...
+
+                <div class="flex items-center space-x-2 leading-tight">
+                    <span class="text-white dark:text-theme-secondary-200">Learn more <span class="hidden sm:contents">at MarketSquare</span></span>
```
ItsANameToo: lang file

resources/views/components/marketsquare/banner.blade.php
```diff
@@ -0,0 +1,31 @@
...
+                    <a href="#" class="mx-auto link">
```
ItsANameToo, line 17: add a card that it still needs to be hooked up to an MSQ url if there isn't one yet

faustbrian: https://app.clickup.com/t/96qytq

PR #255, opened 2018-06-02
improves #252 to not add the empty div if no tx has a smartbridge

dated:


ItsANameToo:
:+1:

dated:
i'm not getting this out of my head :D how about filled icon if smartbridge is present, unfilled if not?

image

ItsANameToo:
Hmm, that's also a possibility. Looks cool and solves the marging issues so I'd say go for it :D

dated:
thanks for the feedback!

PR #261, opened 2018-06-04
Disable all html in v-tooltip

ItsANameToo:
@alexbarnsley as talked about on slack.

I've added a default to not allow html anywhere the tooltip is used. If someone really needs to show HTML, then it should be explicitly allowed for that particular instance. Also removed the sanitizeHtml package as we no longer need it.

Before:
schermafbeelding 2018-06-04 om 18 32 09

After:
schermafbeelding 2018-06-04 om 18 31 59

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "lodash": "^4.17.10",
     "moment": "^2.22.1",
     "qrious": "^4.0.2",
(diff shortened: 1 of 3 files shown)


roks0n (approved):
🎉

ItsANameToo:
🎉

PR #267, opened 2018-06-06
splits the timer used in App.vue in two

dated:
The timer used to fetch the network data and exchange rate got fired every five minutes. This PR splits this timer in two, so that the network data gets fetched every 8 seconds. This way the content header component shows the netheight in "realtime".

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -26,7 +26,8 @@ export default {
   components: { AppHeader, AppFooter },
 
   data: () => ({
-    timer: null,
+    currencyTimer: null,
+    networkTimer: null
   }),
 
   async created() {
@@ -91,7 +92,7 @@ export default {
 
   methods: {
     prepareComponent() {
-      this.initialiseTimer()
+      this.initialiseTimers()
     },
 
     async updateCurrencyRate() {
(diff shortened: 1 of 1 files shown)


resources/icons/key.svg
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" x="0" y="0" viewBox="0 0 17.8 8.7" xml:space="preserve"><style>.publicKey-st0{fill:none;stroke:currentColor;stroke-width:1.5;stroke-linecap:round;stroke-linejoin:round}</style><g id="key" transform="translate(-520.75 -361.285)"><g id="car-key-1" transform="translate(552 277.279)"><g id="car-key-1-2" transform="translate(-35.75 84.006)"><circle id="Ellipse_7" class="publicKey-st0" cx="8.9" cy="4.4" r="3.6"/><path id="Path_312" class="publicKey-st0" d="M12.5 4.4h9v2.7"/><path id="Line_9" class="publicKey-st0" d="M17.9 4.4v1.8"/></g></g></g></svg>
```
ItsANameToo, line 1: why not use the frontend repo `key` icon :susp:

Highjhacker: :hide:

Highjhacker: Just looked, it's even worse :joy: And doesn't fit the design ; https://gyazo.com/46e54a50f78e58c36ea383fdb1446475

ItsANameToo: I'll update the frontend key svg tomorrow

PR #287, opened 2018-06-16
Add tooltips

### LINKED ISSUE #286: UX: Add tooltips for truncated addresses
Sometimes it would be helpful to see the complete address by hovering with the mouse. The desktop wallet already does this.

screenshot_20180616_141444

spkjp:
Fixes #286

Small quality of life improvement. Transaction ids and addresses are in most cases truncated. This change adds tooltips.

screenshot_20180616_135256

One thing I tried, but failed to accomplish:
screenshot_20180616_135734

In the transaction detail view the tooltip is superfluous if the address is not truncated. I tried adding some logic tooltip to only display if the address is truncated or isKnown is true.

``` javascript

    content() {
      if (this.walletAddress) {
          
        let truncated = this.truncate(this.walletAddress).indexOf('...') !== -1;
        let known = this.knownWallets.hasOwnProperty(this.walletAddress);
        if (truncated || known) {
          return this.walletAddress;
        }
      }

      return '';
    }
```

But coulnd't get it to work reliably. If it's a blocker I can look at it again or maybe someone with more than an hour of Vue experience can help out.


alexbarnsley:
@supaiku0 did you get chance to take a look at Brian's suggestion?

spkjp:
@alexbarnsley  Sorry, totally forgot about this one. :joy:

Just pushed an update. The tooltips are no longer displayed when not truncated.

alexbarnsley:
@supaiku0 I'm not angry, just disappointed! I will take a look now 👀

alexbarnsley:
Hi @supaiku0, it looks like i'm still getting a tooltip for addresses:

image

If i understood Brian correctly, I think he meant to not show the tooltip for addresses. Probably worth him confirming 🤔

alexbarnsley:
@faustbrian ah ok, I was thinking you meant just delegate names 👌

dated:
I'd like to revisit this, as it introduces a breaking change to transaction tables which have no smartbridge / vendor field column, e.g. https://explorer.ark.io/wallets/ATK14wxyYxbELq2b91bAfBY8Vmh9J6MDej

In this case the value of the smartbridge / vendor field is completely ommitted. We *know* it's there, because of the filled icon, but we cannot access it.

PS: Furthermore I do not see the big advantage of having the full ID / address shown as a tooltip, since there is no way to actually select and copy from the tooltip or use it in any way. As far as identification purposes go, I have yet to come accross two IDs with the same five starting and finishing characters. Having a tooltip for the address of a delegate I fully support.

dated:
Yep, I'll revisit this, since there is even a tooltip when hovering over vote transactions.

image

dated:
I don't want to step on anybodys toes, therefore i kindly request your opinion before changing this, @supaiku0 and @faustbrian.

IMHO there is little to no use in having a tooltip on transaction ids, block ids and addresses. You can't select the text inside, you can't copy it, you can't do anything, other than look at it, probably to identify a transaction. For this use alone, the truncated id, address, whatsoever is enough already.

My suggestions are therefore as follows:

- remove the tooltip on block links
- remove the tooltip showing the id on transaction links and bring back the smartbridge tooltip if there is no smartbridge column
- remove the tooltip on address links, show the address tooltip only when a delegates name is shown

spkjp:
You clearly put a lot of thought into it. :+1: 
It's just a personal preference of mine to have these tooltips, but you raise some valid points.

So feel free to go ahead and make the changes if nobody else is against it, I'm neutral in that regards.

>You can't select the text inside, you can't copy it, you can't do anything, other than look at it, probably to identify a transaction.

This could be improved by adding a clipboard button to the tooltip.

PR #288, opened 2018-06-18
add comma separator to readableNumber function

reConNico:
add comma seperator to readableNumber function and integrate it to forged blocks by delegate

explorer_comma

### DIFF
--- a/src/components/monitor/ActiveDelegates.vue
+++ b/src/components/monitor/ActiveDelegates.vue
@@ -18,7 +18,7 @@
 
     <table-column show="producedblocks" :label="$t('Forged blocks')" header-class="left-header-cell hidden xl:table-cell" cell-class="py-3 px-4 text-left border-none hidden xl:table-cell">
       <template slot-scope="row">
-        {{ row.producedblocks }}
+        {{ readableNumber(row.producedblocks, 0, true) }}
       </template>
     </table-column>
 
--- a/src/mixins/index.js
+++ b/src/mixins/index.js
@@ -91,8 +91,12 @@ const methods = {
       })
   },
 
-  readableNumber(value, digits = 2) {
-    return value.toFixed(digits)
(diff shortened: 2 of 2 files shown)


dated:
I'd use `toLocaleString()` to mantain proper localisation, otherwise :+1:

reConNico:
Integrated `toLocaleString()` as @dated proposed.

ItsANameToo:
@reConNico your current approach won't work, as `.toLocaleString` should be called on a number and `.toFixed` returns a string. Instead, you can call `.toLocaleString` and use its [options parameter](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) to define the amount of digits you want to return. As an example, you could use `value.toLocaleString(undefined, { minimumFractionDigits: digits, maximumFractionDigits: digits })` . In case of two digits, `50100.78` would be turned into `"50.100,78"` for my locale with that function.

reConNico:
@ItsANameToo . Thanks for the tip. Could you review the changes? It wouldn't make sense to also set `maximumFractionDigits` on the same number as `minimumFractionDigits` in my opinion.

ItsANameToo:
In this case you'll also need to set `maximumFractionDigits` to keep the same behaviour as `toFixed()`, as you want to show the number with the specified amount of digits (and no more or less). The `minimumFractionDigits` ensure that numbers such as 100 and 100.1 are formatted as 100.00 and 100.10 if you specified 2 digits, but numbers such as 100.123 will be formatted as 100.123, as you did not specify a `maximumFractionDigits` to keep them at 2. For your forged blocks example it would work, as I don't expect someone to have forged 0.1 blocks, but in the more general sense this function should also make use of `maximumFractionDigits: digits` to keep its original behaviour.

faustbrian:
`maximumFractionDigits` should also be set to keep it consistent with other `toLocaleString` calls.

PR #299, opened 2018-07-08
Dependency updates

ItsANameToo:
Updated all dependencies, except for tailwind (0.5.3 to 0.6.1) as they made some changes that result in things like this:

schermafbeelding 2018-07-08 om 21 46 47

Will take a further look into the exact reason for this.


ItsANameToo:
main concern with estimates: delegates that drop out during and get back in at a later time. Based on the time they would have forged x blocks, but because they were out for y time it won't be close to that so it mistakenly indicates the delegate to have missed blocks

faustbrian:
https://app.clickup.com/t/9rwj1x

Issue #302, opened 2018-07-13
UX: Update Latest Blocks


dated:
**Is your feature request related to a problem? Please describe.**
The Latest Blocks component does not stay up to date.

**Describe the solution you'd like**
Keep fetching the latest blocks when the network height changes.

PR #303, opened 2018-07-13
Refreshes latest blocks #302

dated:


faustbrian:
Could you please add tests for this change.

dated:
An unrelated test, which has been failing sporadically in the latest builds, causes Travis to fail. The test for this PR is looking good.

Issue #304, opened 2018-07-13
UX: Update Latest Transactions


dated:
**Is your feature request related to a problem? Please describe.**
The Latest Transactions component does not stay up to date.

**Describe the solution you'd like**
Keep fetching the latest transactions when the network height changes.

PR #301, opened 2018-07-13
Removes timers and watches store instead #300

dated:
Please see https://docs.ark.io/docs/contributing for details before opening your pull request.

PR #305, opened 2018-07-13
Refreshes latest transactions #304

dated:


faustbrian:
Could you please add tests for this change.

dated:
Testing this will be difficult as there is no guarantee of new transactions for each new block. The fetching of the latest blocks however can be tested. Since both components use the same logic, we should be able to assume that that test validates this change as well. Or maybe you have an idea @ItsANameToo?

ItsANameToo:
I agree that transactions will be difficult. Only possibility I can think of is looking into mocking the `getTransactions()` function and asserting that it gets called when network height changes, but not sure how and if it's possible 😬

dated:
I have not come up with a solution for how to test this, other than indirectly through other tests that confirm the same behaviour. If someone wants to give it a shot, please do.

dated:
Commit b5a795e seems to be missing in master, although being merged with #301

faustbrian:
Merged it for now without tests passing as there seem to be other issues with the test suite.

ItsANameToo:
wip

PR #312, opened 2018-07-30
Fixes delegate status tooltip background color

dated:
Fixes #311

PR #314, opened 2018-07-30
Fixes 'Second Signature Enabled' translation string

### LINKED ISSUE #313: Second Signature translation string
**Describe the bug**
The tooltip when hover over the lock icon on the wallet page is referenced with 'Second Signature Enabled' while it is defined as 'Second Signature' in the english and dutch translation files, resulting in a missing translation warning.

dated:
Fixes #313, and adds Portuguese translation.

### DIFF
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -86,7 +86,7 @@
   "All": "All",
   "Sent": "Sent",
   "Received": "Received",
-  "Second Signature": "Second Signature Enabled",
+  "Second Signature Enabled": "Second Signature Enabled",
 
   "Previous": "Previous",
   "Next": "Next",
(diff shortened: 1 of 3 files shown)


faustbrian (approved):

dated:
I believe the Travis servers are just to slow sometimes @faustbrian. The `retryAssertionTimeout` config added with cd6e77c / #312 should fix this behaviour however, as it will retry the assertion until the timeout is reached..

faustbrian:
What is weird is this error. It has the same values but says they are not equal.

```
   - it should refresh the confirmation count automatically (8.696s)
   Failed [notEqual]: ('608607' != '608607')  - expected "608607" but got: "608607"
```

dated:
The output is weird, I agree. `[notEqual]` is the assertion in this case, which fails because the values are in fact equal. This is more confusing imho: `expected "608607" but got: "608607"`, when it expects anything other than that value :D

faustbrian:
It doesn't make a whole lot of sense and happens randomly. Might be an issue with the timing of getting the data.

PR #317, opened 2018-08-02
fix: Updated OKEx wallet address

boldninja:
Recently OKEx moved their funds to a new wallet, this changes to a new address in Explorer.

### DIFF
--- a/networks/mainnet.json
+++ b/networks/mainnet.json
@@ -30,7 +30,7 @@
     "ALgvTAoz5Vi9easHqBK6aEMKatHb4beCXm": "ARK Shield",
     "AUDud8tvyVZa67p3QY7XPRUTjRGnWQQ9Xv": "ARK Team",
     "AUexKjGtgsSpVzPLs6jNMM6vJ6znEVTQWK": "Bittrex",
-    "AeUyEH2UGpYrwHAupBh7syFhWYSBNFAkap": "OKEx",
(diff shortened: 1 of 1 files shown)


dated:
Hi @boldninja, imho opinion it would still be nice to have the old deprecated address available. There are two ways to go about this:

1) Just include it as previously but add "out of service", "deprecated" or something else to the name
2) Add a separate `knownDeprecatedWallets` or `knownObsoleteWallets` object to the network configuration

The second approach would allow us to add an icon or other visuals along the name to clearly indicate that the wallet is out of service.

Thoughts?

boldninja:
I'm not sure, 

1. would confuse people even more what if they change it back to old one and it says deprecated or out of service for few days until we notice? Would look weird.

2. What would that bring in regards to user experience? 

I think that current model for now is fine, but I'm open for suggestions if others think this would bring up UI value.

dated:
The thing is that we effectively remove information by just replacing the known wallet. A transaction that previously was clearly identifiable as coming from OKEx (or any other known wallet by that matter), suddenly isn't anymore. It's more about retaining the UX rather than adding to it. But i can live happily without it, just my two cents ;)

boldninja:
I'd be more for "OKEx - old" for previous wallet than just saying deprecated or obsolete that can be added to knownObsoleteWallets to distinguish in config and for cleaner config look.

PR #341, opened 2018-08-13
Dependency udpates

dated:
Updated all dependencies. The issue with tailwind described in #299 has been resolved by using the style attribute instead of the width attribute on the image tag (which was being overwritten by the `width: auto;` rule).


dated:
something causes the e2e tests to timeout locally as well. i'll need to investigate this.

faustbrian (approved):

Issue #324, opened 2018-08-07
ARK explorer website (explorer.ark.io) is using tcp/8443 behind the scenes


rdk1979:
**Describe the bug**
If you're behind a corporate firewall that for example only allows HTTP(S) traffic you can't reach the ARK explorer at https://explorer.ark.io.

A netstat -n shows me that my laptop is trying to open TCP/8443 to IP 104.25.220.25.

**To Reproduce**
1. make sure port TCP/8443 is closed on your firewall.
2. Goto https://explorer.ark.io
3. you'll see that "latest transactions" and "latest blocks" keep loading forever. (3 flashing dots)

**Expected behavior**
A website over an secure SSL (tcp/443) connection should preferably not use any other port in the background.

This issue occurs on every system with IE, Edge, Chrome, Firefox, etc.

Neurone:
This is not a real issue of the explorer, it's more a configuration _bother_. API are served by ARK node, not by the explorer, so any node can serve the explorer well. I think the best approach is to allow changing the ARK node to use while surfing the explorer: it should be nice to have an option to change that configuration in the top bar, the same way I change the ARK/currency value, or enable/disable dark mode.

PR #350, opened 2018-08-21
User can now choose a custom API server (see issue #324)

Neurone:
- Setting is saved to local storage for later use
- User can reset to the original value


faustbrian:
Running an explorer on normal nodes can have a major impact on performance, that is why the https://github.com/ArkEcosystem/ark-node/tree/explorer branch exists.

With this feature everyone could just put in any IP and cause performance issues for nodes. I will have to think about merging this.

Neurone:
I didn't get it: you can find the exactly same feature - changing the node you connect to - in the Desktop Wallet and in the Mobile Wallet.

And frankly you scared me a bit with this answer: are you really saying that having this simple feature **at client side** can put any ARK node at risk? If so, it means that **ARK nodes are already at risk of spam attacks**, with or without this feature, and then we have major problems to think about and to solve in the core.

Connecting to an exposed node cannot be considered in any case a bad behaviour: a client should not be able to ever put a node at risk, not because the client acts _fairly_ but because the node must protect itselft by attacks in different ways.

I didn't got a full node up and running yet, but I thought this kind of APIs - let's call them a sort of analytics API - were exposed only by owner's will, didn't they? Core v2 already address this kind of common issues for API exposition (listening only on loopback interface, access control, blacklisting client's considered spammers, throttling, etc.)? If not, I think these issues must be addressed before going live to the Mainnet with the v2.

faustbrian:
The issue is more on the performance side with low-spec servers as the database has millions of records which can put a lot of strain on those servers with low-specs and the explorer produces a whole lot more traffic then a wallet.

All of those issues have been addresses with v2 but currently v1 is on mainnet, not v2.

faustbrian:
I will close this for now and revisit it when v2 is on mainnet. Thanks for the PR though.

Issue #349, opened 2018-08-20
Use local time zone and display time zone used


blmatthews:
Timestamps are displayed in UTC+7. The only way I know that is going to the list of latest transactions (or using a transaction that I know when it took place via some external mechanism) and calculating the offset from my current time, which doesn't seem like something everyone should have to do. So I have two requests:

- Display the time zone used, either next to any timestamp, or at the bottom of the page, or something like that.
- Determine the browser's time zone and use that (and still do the previous).

As a bonus, also display dates using either a localized format or yyyy.mm.dd. Currently it's dd.mm.yyyy, which is ambiguous if the day is <=12, and doesn't sort as easily as yyyy.mm.dd.

dated:
Are you still experiencing this? The date formatting has been tackled in #360 and is now based on your browsers locale.

blmatthews:
I now see the date format as mm.dd.yyyy, but it's still in UTC+7 (I'm UTC-7), and nowhere I can see does it say what time zone it's using, I had to figure it out (so I may be wrong :-) ), so yes, still experiencing this.

dated:
I've identified the issue I think, will do some testing.

ItsANameToo:
@faustbrian plz close

PR #357, opened 2018-08-26
YYYY.MM.DD instead of opposite

tpscrpt:
## Proposed changes

In general, the American date format is only really intuitive to Americans in America. When any internationalization occurs, Americans themselves tend to adapt to the date format quite easily, especially if it's in YYYY.MM.DD format and the following is HH:mm:ss. The only reason I know is because I'm an inbetweener! :canada: (culture here is weird).

This PR just changes the standard getReadableTimestamp mixin to display timestamps from bigger to smaller unit.

#349 raises this issue, probably could use a timezone selection tool regardless.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Build (changes that affect the build system)
- [ ] Docs (documentation only changes)
- [ ] Test (adding missing tests or fixing existing tests)
- [ ] Other... Please describe:

## Checklist

### DIFF
--- a/src/mixins/index.js
+++ b/src/mixins/index.js
@@ -32,7 +32,7 @@ const methods = {
       })
       .add(Math.abs(typeof timeZoneOffset !== 'undefined' ? timeZoneOffset : new Date().getTimezoneOffset()), 'minutes')
       .add(value, 'seconds')
-      .format('DD.MM.YYYY HH:mm:ss')
+      .format('YYYY.MM.DD HH:mm:ss')
   },
 
   readableTimestampAgo(time, compareTime) {


dated:
You can use `.format('L LTS')` to get a localized timestamp, see [https://momentjs.com/#multiple-locale-support](https://momentjs.com/#multiple-locale-support). But the language always falls back to english apparently. The App tries to get the language from `localStorage('language')` which is never set in the language setter method of the [ui store](https://github.com/ArkEcosystem/explorer/blob/master/src/store/modules/ui.js#L30).

``` javascript
# App.vue

this.$store.dispatch(
  'ui/setLanguage',
  localStorage.getItem('language') || 'en'
)
```

 You can probably use `navigator.language` to get the proper localisation, which has to be set in moment through `moment.locale(store.getters['ui/language'])`.

tpscrpt:
@dated I'm not a vue developer and not familiar with moment, but will give it a try.

faustbrian:
As @dated said it is better to localise the timestamp if we already change it rather then just using the American format which is not used anywhere else.

tpscrpt:
not interested in fixing this anymore, my apologies. Still do think that the biggest to lowest unit approach is better than the American default

PR #366, opened 2018-09-11
fix: use total to correctly calculate voter weight on mobile

ItsANameToo:
## Proposed changes

On mobile the "supply" of a voter was based on the total supply of the network, while on desktop it was based on the total amount of votes for a delegate. This PR changes the mobile version to also calculate with the total amount of votes for a delegate to show the same percentage

## Types of changes

## Checklist

### DIFF
--- a/src/components/tables/mobile/Wallets.vue
+++ b/src/components/tables/mobile/Wallets.vue
@@ -19,7 +19,7 @@
 
         <div class="list-row">
           <div>{{ $t("Supply") }}</div>
-          <div>{{ readableNumber((row.balance / supply) * 100) }}%</div>
+          <div>{{ readableNumber((row.balance / total) * 100) }}%</div>
         </div>
       </div>
       <div v-if="wallets && !wallets.length" class="px-5 md:px-10">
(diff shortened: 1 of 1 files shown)


resources/views/components/grid/generic.blade.php
```diff
@@ -1,5 +1,7 @@
 @if(isset($withoutBorder) && $withoutBorder === true)
 <div class="flex items-center justify-between pb-4">
+@elseif(isset($responsiveBorder) && $responsiveBorder === true)
+<div class="flex items-center justify-between pb-4 border-b border-theme-secondary-300 dark:border-theme-secondary-800 md:border-none">
```
ItsANameToo: isn't this just copy paste of the below line but with `md:border-none` added?

Highjhacker: It is, how'd you do differently ? 🤔

ItsANameToo: then add it to the bottom line and set `md:border-none` if needed, not copy the whole line 🤔

PR #368, opened 2018-09-11
feat: add tooltip for smartbridge message

ItsANameToo:
## Proposed changes

A while ago it was possible to see the smartbridge message in a tooltip if a transaction had one by hovering over the id. This has now been replaced with showing the full id, but I miss the quick look at a smartbridge message, so this PR adds back a tooltip to show it. The tooltip is only shown when hovering over the "label" icon of a transaction. On mobile the table already shows the smartbridge message if one is available.

## Types of changes

## Checklist

### DIFF
--- a/src/components/links/Transaction.vue
+++ b/src/components/links/Transaction.vue
@@ -1,10 +1,12 @@
 <template>
-  <span v-tooltip="id">
-    <router-link :to="{ name: 'transaction', params: { id } }" class="flex items-center whitespace-no-wrap">
-      <span v-if="hasDefaultSlot"><slot></slot></span>
-      <template v-else>
+  <router-link :to="{ name: 'transaction', params: { id } }" class="flex items-center whitespace-no-wrap">
+    <span v-if="hasDefaultSlot"><slot></slot></span>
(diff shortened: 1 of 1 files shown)


dated:
Welcome back! There has been a bit of discussion around this, see #287

ItsANameToo:
@dated thanks! Hadn't seen that discussion, so thanks for pointing it out to me. 

This PR is probably not the best solution seeing the options that you've come up with in the comments there. I also don't see the use of the transaction ID tooltip, but if some people find it useful I'm fine with keeping it though. Anyway, if you want to implement your suggestions then let me know and I'll close this PR. As long as I can see the smartbridge text again without the need of clicking through to the transaction detail I'm happy 😬

ItsANameToo:
@faustbrian don't merge this yet

dated:
The missing smartbridge tooltip was my _main_ concern, i am basically fine with whatever solution brings that back and this does seem like a good one 👍

ItsANameToo:
@faustbrian you can merge it then 🤔

faustbrian (approved):

PR #369, opened 2018-09-11
chore: add Binance cold wallet to known wallets

ItsANameToo:
## Proposed changes

Adds Binance cold wallet to the list of known wallets (it's the wallet they move all their funds to and occasionally put ARK out when they are near 0 in their regular wallet)

## Types of changes

- [x] Other... Please describe: chore

## Checklist

### DIFF
--- a/networks/mainnet.json
+++ b/networks/mainnet.json
@@ -24,6 +24,7 @@
   },
   "knownWallets": {
     "AFrPtEmzu6wdVpa2CnRDEKGQQMWgq8nE9V": "Binance",
+    "AQkyi31gUbLuFp7ArgH9hUCewg22TkxWpk": "Binance Cold Wallet",
     "AJbmGnDAx9y91MQCDApyaqZhn6fBvYX9iJ": "Cryptopia",
     "AYCTHSZionfGoQsRnv5gECEuFWcZXS38gs": "ARK Bounty",
     "AZmQJ2P9xg5j6VPZWjcTzWDD4w7Qww2KGX": "ARK GitHub Bounty",


resources/views/components/home/tabs.blade.php
```diff
@@ -1,6 +1,6 @@
 <div x-data="{
     tabsOpen: false,
-    selected: 'transactions',
+    selected: 'blocks',
```
ItsANameToo: sus

resources/views/components/tables/mobile/blocks.blade.php
```diff
@@ -7,8 +7,6 @@
 
             <x-tables.rows.mobile.block-forger :model="$block" />
 
-            <x-tables.rows.mobile.block-height :model="$block" />
```
ItsANameToo: why do we remove data?

resources/views/components/tables/rows/mobile/transaction-count.blade.php
```diff
@@ -1,5 +1,7 @@
 <div>
     @lang('labels.transaction_count')
 
-    {{ $model->transactionCount() }}
+    <div class="inline">
```
ItsANameToo: use a `span` if you want something `inline`

resources/views/components/home/tabs.blade.php
Highjhacker: Oh yeah forgot that one

resources/views/components/tables/mobile/blocks.blade.php
Highjhacker: Not in the designs

ItsANameToo: then design is off

Highjhacker: Ah, was mentioned in the card too by @samharperpittam but will add back then :vinigotit:

PR #394, opened 2018-10-01
hardens transactions recipient test

dated:
## Proposed changes

The original 'recipient test' would fail if the first transaction was not a simple transfer, i.e. it had no link in the recipient column. Now the test first checks if such a link is present, and continues only with the test logic if that's the case.

## Types of changes

## Checklist


faustbrian:
If you have more tests that need to be adjusted please include them in this PR.

dated:
That'd be all for the time being :)

faustbrian (approved):

PR #427, opened 2018-10-29
chore: update nightwatch

dated:
## Proposed changes

This PR updates the Nightwatch version as there are several vulnerabilities

```bash
found 6 vulnerabilities (1 low, 4 high, 1 critical)
```

## Types of changes

- [x] Other... Please describe: security updates

## Checklist


faustbrian (approved):

marianogoldman:
@faustbrian This PR I guess it's ready to be merged as is. The only thing it does is to defer all possible user updates in UI to the time they click `Find It`. The search type (transaction, block, wallet) cannot be deferred because that causes the form to change so that other input fields are displayed.

What this PR does NOT include is the change we were discussing with Sam / Mathew on Slack earlier today. I tried to implement it but with things as they are I don't think it's possible, because... given that search type still needs to be dynamic it will trigger an update, and by then we don't have the search results we already rendered to use them for re-render. I also doesn't make sense to re-execute the currently displayed search again just to keep the UI from changing. I tried making `$results` a public property (so it would get rehydrated in each cycle, at the cost of larger payloads for updates), but it is a `LengthAwarePaginator`, so it can't be handled as a public property. So we definitely need to run the search again. I attempted something here: https://github.com/ArkEcosystem/explorer.ark.io/pull/440, and realized it was not possible.

So I guess we have 2 options:

* Not do what we said was the best, and use the search as a live filter that reacts to changes in search type only or clicks in Find It.

* Implement the change of search type in the front end (Alpine.js), so that the form is rendered complete, with hidden parts. When search type is selected, we dynamically update the form and then finally, only on submission (Find It) we trigger a new search in the backend.

Unless you have a better idea... let me know.

faustbrian:
@marianogoldman use Alpine, that's the easiest to avoid a re-render of the component.

PR #435, opened 2018-11-03
refactor: remove whitespace from search param

dated:
## Proposed changes

Trims query search param.

## Types of changes

## Checklist

### DIFF
--- a/src/components/header/Search.vue
+++ b/src/components/header/Search.vue
@@ -55,6 +55,7 @@ export default {
     async search() {
       this.nothingFound = false
       this.searchCount = 0
+      this.query = this.query.trim()
 
       try {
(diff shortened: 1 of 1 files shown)


faustbrian (approved):

resources/views/components/icon.blade.php
```diff
@@ -20,4 +20,4 @@
     }
 @endphp
 
-@svg($name, $size.($style ?? '').($class ?? ''))
+@svg($name, $size . ' ' . ($style ?? '') . ($class ?? ''))
```
ItsANameToo: same issue will occur when a `style` is set

faustbrian: Why not modify the `$style` and `$class` properties to already contain the spacing they need?

PR #439, opened 2018-11-05
setting correct flag for brazil

Pedro-Souza:
## Proposed changes

setting correct flag for brazil

## Types of changes

## Checklist


src/i18n/index.js
```diff
@@ -5,10 +5,10 @@ import store from '@/store'
 Vue.use(VueI18n)
 
 const messages = {
-  en: require('./en.json'),
-  nl: require('./nl.json'),
-  pt: require('./pt-br.json'),
-  pl: require('./pl.json')
+  'en': require('./en.json'),
+  'nl': require('./nl.json'),
```
faustbrian: No need to escape since it has no hyphens.

src/i18n/index.js
```diff
@@ -5,10 +5,10 @@ import store from '@/store'
 Vue.use(VueI18n)
 
 const messages = {
-  en: require('./en.json'),
-  nl: require('./nl.json'),
-  pt: require('./pt-br.json'),
-  pl: require('./pl.json')
+  'en': require('./en.json'),
```
faustbrian: No need to escape since it has no hyphens.

src/i18n/index.js
```diff
@@ -5,10 +5,10 @@ import store from '@/store'
...
+  'nl': require('./nl.json'),
+  'pt-br': require('./pt-br.json'),
+  'pl': require('./pl.json')
```
faustbrian: No need to escape since it has no hyphens.

Pedro-Souza:
Done @faustbrian

dated:
The failing test needs to be fixed here:

https://github.com/ArkEcosystem/explorer/blob/master/test/unit/specs/i18n/i18n.spec.js#L11

`expect(i18n.t('No Results', 'pt-br')).toBe('Sem resultados')`

Pedro-Souza:
Done @dated @faustbrian

dated:
I'll be looking for an icon set for the flags, after that this is ready to merge. 17.1 KB for a flag is simply too much.

dated (approved):

faustbrian:
Mark this explicitly as a development command in the docs because this should not be used in production cronjobs. Ideally we would only mention it in the development docs, not the general cache docs.

PR #440, opened 2018-11-06
refactor: add hover trigger to approval tooltip

dated:
## Proposed changes

Adds click trigger to the approval tooltip, so that it can be accessed on mobile.

## Types of changes

## Checklist


faustbrian (approved):

app/Http/Livewire/SearchPage.php
```diff
@@ -27,6 +27,8 @@ final class SearchPage extends Component
 
     private ?LengthAwarePaginator $results = null;
```
marianogoldman, line 28: @faustbrian This property cannot be made public (and hence have LiveWire rehydrate it on each request, at the cost of a larger payload off course). So we have no other option than to run the search again. Which I think it's a waste of time, given we still expect the user to enter other details in other fields before the search makes sense.

resources/views/livewire/search-page.blade.php
```diff
@@ -23,15 +23,15 @@
 
             @if($results && $results->count())
                 <div>
-                    @if ($state['type'] === 'block')
```
marianogoldman, line 26: This was an attempt to re-render not based on current selection (which is being updated), but on latest search results. Of course it doesn't work, because `$results` didn't keep state.

PR #447, opened 2018-11-07
Add missing translation in delegate monitor

bbrtj:
## Proposed changes
There was a translation missing from tooltip in delegate monitor. Text was always displayed in English (no i18n call).
I've added a call to i18n interface and filled translations for English (same as it was) and Polish as these are the languages I speak. I've also added placeholders in every other translations file in English, so this will remain unchanged. Someone who speaks one of these two languages should translate.

## Types of changes

## Checklist


bbrtj:
Also one unit test is failing, which doesn't seem to be connected at all to my changes

dated (approved):
I'm fixing the unit test in a different PR, will merge here when ready.

PR #453, opened 2018-11-09
refactor: add support for different locales of the same language

dated:
## Proposed changes

Adds support for multiple locales of the same language and allows to differentiate between british and american english. While the same 'translations' are used, the locale and formatting of dates, times etc. is drastically different.

## Types of changes

## Checklist

## Further comments

Somehow timestamps are not updated immediately when switching between both english versions, this is noticeable on the latest blocks or block details table for instance. It does however work as expected with every other language 🤔


dated:
Anyhow, since setting the language should be a one time task i don't see much of a problem in it.

faustbrian (approved):

resources/views/components/home/charts.blade.php
```diff
@@ -29,7 +29,7 @@ class="content-container"
             />
         </div>
 
-        <div class="grid w-full grid-flow-row grid-cols-2 gap-6 pt-8 mt-5 mb-16 {{ Network::canbeExchanged() ?  'xl:grid-cols-4' : 'xl:grid-cols-3'}} gap-y-12 xl:gap-y-4">
+        <div class="grid w-full grid-flow-row grid-cols-2 gap-6 pt-8 mt-5 mb-16 {{ Network::canbeExchanged() ?  'xl:grid-cols-4' : 'xl:grid-cols-3'}} gap-y-12 xl:gap-y-4 chart-details chart-details">
```
ItsANameToo: `chart-details chart-details` :sus:

resources/css/_general.css
```diff
@@ -253,18 +253,34 @@ body {
     .delegate-details > .detail-box:nth-child(2n + 1) {
         @apply border-r pr-6 border-theme-secondary-300;
     }
+
+    .chart-details > .detail-box:nth-child(2n + 1) {
+        @apply border-r pr-6 border-theme-secondary-300;
```
ItsANameToo, line 260: if lines are identical to other css rules, just add them together

resources/css/_general.css
```diff
@@ -253,18 +253,34 @@ body {
...
+    }
 }
 
 @screen lg {
     .delegate-details > .detail-box:not(:last-child) {
         @apply border-r pr-6 border-theme-secondary-300;
     }
+
+    .chart-details > .detail-box:not(:last-child) {
+        @apply border-r pr-6 border-theme-secondary-300;
```
ItsANameToo, line 272: same as above

resources/css/_general.css
```diff
@@ -253,18 +253,34 @@ body {
...
+    }
 }
 
 .theme-dark .delegate-details > .detail-box {
     @apply border-theme-secondary-800;
 }
 
+.theme-dark .chart-details > .detail-box {
+    @apply border-theme-secondary-800;
```
ItsANameToo, line 282: same as above

ItsANameToo:
I meant combining in the sense of doing this:

```
.delegate-details > .detail-box:nth-child(2n + 1),
.chart-details > .detail-box:nth-child(2n + 1) {
    ...css rules here
}
```
but renaming works too I suppose. I don't like the name for it tho :danstop:

ItsANameToo:
needs a few more changes so I'll update it

PR #456, opened 2018-11-11
adding close icon when menu is open

Pedro-Souza:
## Proposed changes

I think it's better to have an icone to close when the menu is open, the way it was a little confused
captura de tela de 2018-11-11 20-48-58

## Types of changes

## Checklist

### DIFF
--- a/src/components/header/menu/Desktop.vue
+++ b/src/components/header/menu/Desktop.vue
@@ -1,17 +1,9 @@
 <template>
   <div class="absolute pin px-4 md:px-8 hidden sm:flex bg-theme-nav-background xl:rounded-r-md">
-    <button
-      @click="closeMenu"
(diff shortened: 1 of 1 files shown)


dated:
I'm pretty sure that the icon is like that by design.

faustbrian:
This is the intended design.

resources/views/components/search.blade.php
```diff
@@ -163,7 +168,9 @@ class="w-full dark:text-theme-secondary-200 dark:bg-theme-secondary-900"
                             />
                         </div>
                     </x-general.search.advanced-option>
+                </template>
 
+                <template x-if="searchType === 'block'">
```
marianogoldman, line 173: @faustbrian I know I'm repeating this <template x-if> element a lot. But... <template> with <v-if> can only contain one element inside (not many). And if I add a <div> I am breaking the layout and all fields get into a single column. I don't know if there is any other way around it, but happy to change it if there's a better way.

PR #451, opened 2018-11-08
chore: translation cleanup

dated:
## Proposed changes

Cleans up the translation files.

## Types of changes

- [x] Other... Please describe: translations

## Checklist

- [x] I have read the [CONTRIBUTING](https://docs.ark.io/guidebook/contribution-guidelines/contributing.html) documentation
- [x] Lint and unit tests pass locally with my changes

## Further comments

The usual delegate search test has failed in the circleci build.

PR #469, opened 2018-11-20
Key and values fixes, wasn't translated at all before that because of a key mismatch

Highjhacker:
## Proposed changes
The words "Public Key" and "Wallet Summary" weren't translate properly because of a key mismatch, fixed it.

## Types of changes

## Checklist


dated (changes requested):
the lowercased strings were introduced in #451, where i apparently failed to change some occurrences of `Wallet Summary` into `Wallet summary` (and the same for public key). So that is what should be done instead of capitalizing them again :D if you want to take care of that, then cool.

dated (approved):

PR #476, opened 2018-11-22
refactor: more reliable clipboard tooltip

dated:
## Proposed changes

The tooltip of the clipboard button did not work as intended, causing the `Copied!` message to disappear almost immediately.

clipboard_old

This PR refactors the Clipboard component and adds a background color based on the success.

clipboard_new_success clipboard_new_error

## Types of changes

## Checklist

### DIFF
--- a/src/components/utils/Clipboard.vue
+++ b/src/components/utils/Clipboard.vue
@@ -1,8 +1,6 @@
 <template>
-  <button class="flex-none" @click="copy" v-tooltip="{ show: isMobileCopying, content: copying ? $t('Copied!') : $t('Copy to clipboard'), trigger:'hover'}">
-    <img :class="{
-      'block': !copying, 'block animated wobble': copying
-    }" src="@/assets/images/icons/copy.svg" ref="copyImage" />
+  <button class="flex-none" @click="copy" v-tooltip="getTooltip()">
+    <img class="block" :class="{ 'animated wobble': copying }" src="@/assets/images/icons/copy.svg" ref="copyImage" />
   </button>
 </template>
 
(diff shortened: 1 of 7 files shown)


faustbrian (approved):

ItsANameToo:
minor comments:
- the hover should have less margin at the top / bottom to match the design more closely
- the monitor needs a similar styling (e.g. when blocks are missed, see screenshot below), but this is not done "on hover" as it should remain on screen at all times. Is that something we could achieve with the js solution too? otherwise this solution won't solve all our problems

image

alfonsobries:
@ItsANameToo I think is possible with pure CSS but the only way I think is to always add an extra column at first and at the end maybe with a span inside to emulate the rounded corners, that will also require some custom CSS I don't think it will be achievable with pure tailwind, the problem is that in every table we will need to add columns, spans, classes etc.

The JS alternative is similar to the one for the hover effect but thinking of adding some attribute to the rows like `<tr data-warning>` then with JS clone a span with the effect beside the rows according to the attribute. I don't super love to add random `spans` with JS but this alternative gives us more flexibility with the design and opens the possibility to make some reusable component, so unless you have a better opinion I think I will try this option.

ItsANameToo:
adding DOM elements means it's not possible with pure CSS :trollface:

alfonsobries:
@ItsANameToo I'm pausing this over more priority tasks but just pushed a proof of concept by hardcoding the `data-warning` and `data-danger` attributes to the rows, the table looks like this

Still, there are a few issues I need to take care
- When the table is live updated it remove the classes
- If the browser is resized

image

alfonsobries:
@ItsANameToo wake up thinking that we should consider using grids instead of tables

alfonsobries:
@ItsANameToo so just pushed a working POC of the same effect using pure CSS

Means:

1. The cells will need to be added as a component (to ensure they have the correct classes) or:
2. Alternatively we can just define all those CSS classes inside the `laravel-ui/resources/assets/css/_tables.css` file, I think the second option is fine since we are defining the table classes in the CSS file anyway.

Let me know what do you think?

faustbrian:
Screenshot 2020-11-11 at 7 42 57

alfonsobries:
@ItsANameToo 
~~Update: this branch now depends on this PR https://github.com/ArkEcosystem/laravel-ui/pull/61 and I will use it to update all the tables in the explorer to use the new cell components, meanwhile it has a working example that "emulates" warning and error rows in case you want to test it while I finish~~

Update 2: Is ready to review

ItsANameToo:
@alfonsobries conflicts, and you can update to latest laravel-ui dep to get your whitelist changes

alfonsobries:
Once this is merged (https://github.com/ArkEcosystem/laravel-ui/pull/68) I can update the dependencies and this should be good to go

PR #488, opened 2018-12-07
refactor: core 2.0 api

dated:
## Proposed changes

Implements the Core 2.0 API endpoints.

On hold until ArkEcosystem/core#1625 has been merged into the core repository, as it depends on the following commits:

- [refactor(core-p2p): add height to the last block of a delegate](https://github.com/ArkEcosystem/core/pull/1625/commits/9997ce93a8aa1a6ddde5acef8fc170ab6b0d3598)
- [fix(core-api): create valid sub-resource cache keys](https://github.com/ArkEcosystem/core/pull/1625/commits/1363afdc25d35dd3bcacaafe061102824633025f)

## Types of changes

## Checklist

## Further comments

The support for the `getNextForgers`, which has been used by the delegate monitor, was dropped in the Core 2.0 API due to security reasons. The refactored delegate monitor will only show the following three statuses for the same reasons:

- Forging
- Missing
- Not Forging

The `Awaiting Slot` statuses have been dropped.


src/components/block/Transactions.vue
```diff
@@ -32,16 +32,23 @@ export default {
 
   watch: {
     block() {
+      this.resettTransactions()
       this.getTransactions()
     }
   },
 
   methods: {
+    resettTransactions() {
```
faustbrian: `resetTransactions `

src/components/block/Transactions.vue
```diff
@@ -32,16 +32,23 @@ export default {
 
   watch: {
     block() {
+      this.resettTransactions()
```
faustbrian: `resetTransactions `

PR #491, opened 2018-12-09
fix: set default language to "en"

ItsANameToo:
## Proposed changes

Should fix the issues with certain texts not being translated properly, as we don't have an `en-gb.json` file


ItsANameToo:
nevermind, doesn't fix the issue as it should be 'en-gb': https://github.com/ArkEcosystem/explorer/blob/master/src/i18n/index.js#L8

PR #492, opened 2018-12-09
fix: pass lang along as it wasn't reliably updated

ItsANameToo:
## Proposed changes

Can't really test it as it worked locally before :trollface:

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -65,23 +65,18 @@ export default {
     this.$store.dispatch('network/setSymbol', response.symbol)
     this.$store.dispatch('network/setNethash', response.nethash)
 
-    this.$store.dispatch(
-      'ui/setLanguage',
-      localStorage.getItem('language') || 'en-gb'
-    )
-
-    this.$store.dispatch(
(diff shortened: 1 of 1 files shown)


ItsANameToo:
seems to be a caching issue as clearing cache for the explorer website properly populates these values 🤷‍♂️

ItsANameToo:
@samharperpittam @faustbrian : simple solution to have a hover state for tables, but some issues:

- it's not 100% like the design, it has no top/bottom margin and uses scaling to get the overflow on the sides (like the old explorer)
- on most tables, hover is slow. I suspect it's due to the livewire update stuff as the hover works fine when you go to the wallet page. Homepage / monitor are all sluggish and the hover there makes it look terrible, but I want someone else to test it too so I'm sure it's not my chrome
- for the actual design implementation we'll have to refactor the tables, but not sure if we want to do that for MVP

ItsANameToo:
draft because no dark mode at the moment

ItsANameToo:
closing for the time being

PR #494, opened 2018-12-09
add tooltip with amout in transactionAmount

Pedro-Souza:
## Proposed changes

adding tooltip with total

## Types of changes

## Checklist


dated:
This doesn't add valuable information as the shown amount doesn't take the exchange rate into account. Since you would need to fetch the rate for each shown transaction this had been left out on purpose. Showing the converted value on the details page is enough imho.

PR #500, opened 2018-12-15
misc: remove DARK/ARK when on devnet

kalgoop:
## Proposed changes

For the Devnet, Instead of DARK/ARK : 1 in mobile view which doesn't mean anything, Network : Development is displayed.
On Mainnet, it hasn't been changed.

Also, instead of ==, === have been used to compare the alias(network) in ContentHeader.vue

## Types of changes

## Checklist


src/components/ContentHeader.vue
```diff
@@ -5,16 +5,20 @@
...
+        <div v-if="alias === 'Main'">{{ $t("Market Cap") }}: <currency :amount="+supply"></currency></div>
       </div>
     </div>
     <div class="sm:hidden flex items-center justify-between text-theme-text-tertiary text-2xs px-5 sm:px-8 xl:px-6 py-3 bg-stat-background">
       <div class="mr-2">
         <span>{{ $t("Height") }}:</span>
         <span class="block md:inline-block">{{ height.toLocaleString() }}</span>
       </div>
-      <div class="mr-2">
+      <div class="mr-2" v-if="alias !== 'Main'">
```
faustbrian: Why would we add a conditional to showing the network alias?

kalgoop: Mainnet is the default, in my opinion. And it need not be shown to the user.

PR #501, opened 2018-12-15
fix: null value of ARK/DARK symbol

kalgoop:
readableCrypto function returns "" instead of NULL for the network(ark/dark) symbol.
This fixes the null values shown while the page is loading. null is visible very often in case of slow networks

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactoring (improve a current implementation without adding a new feature or fixing a bug)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Build (changes that affect the build system)
- [ ] Docs (documentation only changes)
- [ ] Test (adding missing tests or fixing existing tests)
- [ ] Other... Please describe:

## Checklist

- [ ] I have read the [CONTRIBUTING](https://docs.ark.io/guidebook/contribution-guidelines/contributing.html) documentation


kalgoop:
ezgif com-video-to-gif
^ explorer.ark.io

It loads as follows-
0 null -> 0 A -> 138,389,238 A

The same happens wherever an amount loads (for example transaction table, also loads like the above).

src/mixins/index.js
```diff
@@ -147,7 +147,7 @@ const methods = {
         maximumFractionDigits: decimals,
       })
 
-      return appendCurrency ? `${value} ${store.getters['network/symbol']}` : value
+      return appendCurrency ? `${value} ${ store.getters['network/symbol'] || "" }` : value
```
dated:
```suggestion
      return appendCurrency ? `${value} ${
          store.getters['network/symbol'] ||
          store.getters['network/defaults'].symbol ||
          ""
        }` : value
```

This would be a better approach imho. You'd need to add the symbol of the respective network to the network files, e.g. https://github.com/ArkEcosystem/explorer/blob/master/networks/mainnet.json#L43

dated (changes requested):

kalgoop:
@dated forgot to apply your suggestion😅

dated (dismissed):

dated (approved):
This works as is but requires some refactoring which I'm going to submit in it's own PR @faustbrian.

PR #502, opened 2018-12-16
fix: disappearing closeSearch button

kalgoop:
screen shot 2018-12-16 at 19 05 39

This is iPhone 6/7/8 viewport. Notice that there is no close button for search. This was due to the shrinking of close button due to flexbox shrinking of the button.
## Proposed changes

flex-shrink has been overridden by the stype properties. 

## Types of changes

## Checklist


dated (changes requested):
Good catch, unfortunately adding `flex-shrink: 0` alters the size of the icon (the tailwind class `flex-no-shrink` does exactly that btw).

kalgoop:
@dated So I'm now using tailwind css class flex-none to prevent growing or shrinking. I have checked on various viewports and everything seems to be working fine.

src/components/header/Search.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="w-full flex items-center px-5 sm:px-10">
-    <button @click="$store.dispatch('ui/setHeaderType', null)">
+    <button @click="$store.dispatch('ui/setHeaderType', null)" class="flex-none">
```
dated:
```suggestion
    <button @click="$store.dispatch('ui/setHeaderType', null)" class="flex flex-no-shrink">
```

I'd do it like this, what do you think?

kalgoop: I don't think you need to apply flex class again. 
And flex-none and flex-no-shrink imply the same thing here. flex-none restricts grow also. However, here svg won't grow though, so they both mean same

dated: can't test right now, but i think without `flex` the icon won't be vertically centered.

kalgoop: Yup. you are right. Visually that doesn't make much difference. But yeah it'll be better.

dated (approved):

app/Services/Search/Concerns/FiltersMultiPaymentValueRange.php
```diff
@@ -0,0 +1,46 @@
...
+     */
+    private function queryMultiPaymentValueRange(Builder $query, $from, $to, bool $useSatoshi = true): Builder
+    {
+        if (is_null($from) && is_null($to)) {
+            return $query;
+        }
+
+        $query->where('amount', '=', 0);
+
+        $query->whereExists(function (\Illuminate\Database\Query\Builder $qq) use ($useSatoshi, $to, $from): void {
```
marianogoldman: Hey @faustbrian, it took me a long while to get this to work. But it's (almost) here. It works fine when looking for multi-payment explicitly, I still need to deal when looking by amount but not specifying transaction type (i.e. all).
The code is messy and not very nice, but want to get the tests to pass first and the refactor it as needed.

tests/Unit/Services/Search/TransactionSearchTest.php
```diff
@@ -118,7 +120,30 @@
     expect($result->get())->toHaveCount(10);
 });
 
-it('should search for blocks by fee minimum', function () {
+it('should search for multipayment transactions by amount range', function () {
```
marianogoldman, line 124: This test is triggering the search correctly, but I get an error 
```
  SQLSTATE[42883]: Undefined function: 7 ERROR:  function jsonb_array_elements(json) does not exist
LINE 1: ..." = $3 and exists (select i.id from ( SELECT id, (jsonb_arra...
                                                             ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts. (SQL: select * from "transactions" where "type_group" = 1 and "type" = 6 and "amount" = 0 and exists (select i.id from ( SELECT id, (jsonb_array_elements(asset -> 'payments') ->> 'amount')::bigint am FROM transactions t WHERE t.id = id ) i where i.id = transactions.id group by "i"."id" having sum(am) >= 90000000000 and sum(am) <= 110000000000))
```
See here: https://github.com/ArkEcosystem/explorer.ark.io/pull/502/checks?check_run_id=1361254987#step:11:1565

Not sure what's going on. The same happens when I run it locally, so not sure what it can be as I'm using the same db server (different databases) for the dev site and for tests. I need to get this working tomorrow. Then I can work on refactors in the above code.

PR #535, opened 2019-01-03
feat: Implemented transactions filter on the homepage

Highjhacker:
## Proposed changes

New feature similar to the one I recently pushed. This one allow to filter  the transactions from the homepage of the explorer and will set the choice of the user in the localStorage, that way if user click on "Show more" he will continue to see the filter type he chosed, and vice versa from the transactions page to the homepage. Also added tests for that new feature. Also added the 'Transfer' translation key who were missing.

Note : The filtering on the homepage is quite slow, take a few seconds, so maybe something not done properly on my side, let me know and I will correct.

## Types of changes

## Checklist


test/e2e/specs/homepage.js
```diff
@@ -398,5 +398,18 @@ module.exports = {
       .useXpath()
       .waitForElementVisible("//div[contains(@class, 'tooltip-inner') and text() = 'Nothing matched your search']")
     browser.end()
+  },
+
+  'it should contains a dropdown allowing to filter transactions types': function (browser) {
+    browser
+      .assert.urlContains('/#')
```
ItsANameToo, line 404: don't do this, the `#` won't be there. Just remove the assert line altogether

ItsANameToo: ok my bad, it does use the `#` but your test will fail since the previous test calls `browser.end()` to close the browser. So either remove it from there and add it to this test, or start the browser again on a specified URL

ItsANameToo (changes requested):

faustbrian (approved):

app/Http/Livewire/WalletQrCode.php
```diff
@@ -50,8 +50,8 @@ public function getWalletUriProperty(): string
     {
         $data = [
             'recipient'   => $this->address,
-            'amount'      => $this->amount,
-            'vendorField' => $this->smartbridge,
+            'amount'      => $this->amount ?: null,
+            'vendorField' => $this->smartbridge ?: null,
```
faustbrian: Isn't this null by default?

app/Http/Livewire/WalletQrCode.php
```diff
@@ -50,8 +50,8 @@ public function getWalletUriProperty(): string
     {
         $data = [
             'recipient'   => $this->address,
-            'amount'      => $this->amount,
-            'vendorField' => $this->smartbridge,
+            'amount'      => $this->amount ?: null,
```
faustbrian: Isn't this null by default?

alexbarnsley: Yeh but if you change the field and then empty it, it goes to an empty string. Unless you have a better approach 🤔

app/Http/Livewire/WalletQrCode.php
alexbarnsley: Yeh but if you change the field and then empty it, it goes to an empty string. Unless you have a better approach 🤔

app/Http/Livewire/WalletQrCode.php
```diff
@@ -34,15 +44,30 @@ public function updated(string $propertyName): void
     public function toggleQrCode(): void
     {
         $this->isOpen = ! $this->isOpen;
+
+        if ($this->isOpen) {
```
ItsANameToo, line 51: wouldn't this be better to clear when closing? what happens if you open the form, fill in values and then close it? will they still be used in the QR that gets generated?

app/Http/Livewire/WalletQrCode.php
faustbrian: Na then it's fine.

faustbrian:
@alexbarnsley conflicts

app/Http/Livewire/WalletQrCode.php
alexbarnsley: nah, it doesn't really matter which way it's done. the qr code would be updated as soon as those values change and get reset. I can do it on close if that would make you more comfortable :bleh:

alexbarnsley:
Sorry, missed the conflict here @faustbrian

ItsANameToo:
@alexbarnsley fix the phpstan issues please

alexbarnsley:
Fixed @ItsANameToo

PR #536, opened 2019-01-03
fix: make tests pass again and other circle issues

ItsANameToo:
## Proposed changes

Got tired of the tests failing all the time, so this should fix it. One thing to note is that the confirmations are no longer automatically updated on the explorer, so for now I commented that test as it will never pass.

## Types of changes

## Checklist


dated:
Is there a rough ETA on the typescript migration on mainnet @faustbrian?

Highjhacker:
Seems good to me 🤔 Can't find any other external links like that

PR #538, opened 2019-01-04
refactor: display default token while waiting for network token

dated:
## Proposed changes

Similar to https://github.com/ArkEcosystem/explorer/pull/501, this adds a fallback value for the network token while it is being fetched.

## Types of changes

## Checklist


package.json
```diff
@@ -125,8 +125,8 @@
     "webpack-merge": "4.1.4"
   },
   "engines": {
-    "node": ">= 6.0.0",
-    "npm": ">= 3.0.0"
+    "node": ">= 9.0.0",
```
faustbrian: We should raise this to 10 and drop the 9 tests completely on CircleCI and instead run 10/11.

faustbrian (approved):

resources/views/app/block.blade.php
```diff
@@ -31,7 +31,7 @@
                 <div class="py-16 content-container md:px-8">
                     <div id="transaction-list" class="w-full">
                         <div class="relative flex items-end justify-between mb-8">
-                            <h2 class="text-xl sm:text-2xl">@lang('pages.block.transactions')</h2>
+                            <h2>@lang('pages.block.transactions')</h2>
```
ItsANameToo: you need to pay attention to the classes that were set. `class="text-xl sm:text-2xl"` doesn't translate to `h2` but to `h4` instead, so you should be using `h4` here

resources/views/dashboard.blade.php
```diff
@@ -1,6 +1,6 @@
 <x-app-layout>
```
ItsANameToo: this file can be removed

ItsANameToo:
@samharperpittam can you go over the pages that were changed here to ensure the headings match designs?

PR #542, opened 2019-01-08
refactor: transaction filtering

dated:
## Proposed changes

Redesigns as shown in https://github.com/ArkEcosystem/explorer/issues/541 and refactors/fixes the logic of the transaction filtering on the homepage.

## Types of changes

## Checklist


zillionn:
Haven't looked into your code, but how do you order the txs? I believe currently they're ordered only by timestamp and if they have the same timestamp they rearrange everytime I scroll.

dated:
Imho the API should take care of the correct ordering - and ordering by timestamp does not work correctly there at the moment (ordering by timestamp does return a list that is out of order).

dated:
Timestamp based sorting should be fixed with https://github.com/ArkEcosystem/core/pull/1957

faustbrian (approved):

PR #551, opened 2019-01-16
feat: chart reload button on error

dated:
## Proposed changes

The requests to the CryptoCompare API fail occasionally, causing console errors and without providing feedback to the user. This PR adds error handling and shows a notice if the chart data failed to load correctly.

image

## Types of changes

## Checklist

### DIFF
--- a/src/components/ChartWrapper.vue
+++ b/src/components/ChartWrapper.vue
@@ -1,17 +1,29 @@
 <template>
-  <div :key="componentKey">
-    <div class="flex justify-between items-center px-10 py-8">
-      <h2 class="text-white m-0 text-xl font-normal">{{ $t("Price in") }} {{ currencyName }}</h2>
-      <div>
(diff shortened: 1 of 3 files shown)


faustbrian (approved):

ItsANameToo:
@faustbrian comments regarding these changes, as they are just to understand where calculations might go wrong:

1. Delegate list order seems to be off by 21, so shifting the ranks by 21 before returning it as `activeDelegates` fixes that aspects (question remains how this can be wrong)
2. By changing the order above, the time to forge is incorrectly shown for the newly ordered delegates due to slots still matching the old order (why??), tried to remedy that by shifting the index too but that doesn't really solve it

so tl;dr looks like slots are wrongly calculated? with the same weird offset of 21 as the original `activeDelegates` list

ItsANameToo:
@faustbrian some smaller changes for now that had to be fixed anyway

faustbrian:
@ItsANameToo conflicts

ItsANameToo:
@faustbrian where?

PR #580, opened 2019-03-01
feat: add tooltips to transaction tables with the value, in currency, of amount and fee

j-a-m-l:
## Proposed changes
Add tooltips to display the price, at the moment of reception, of transactions amount and fees.

Since, when navigating to a wallet page is common that each transaction has been received on different dates, this PR includes a mechanism to not surpass the limit of request that the API of Crypto Compare accepts.

## Types of changes

## Checklist


ghost:
@ItsANameToo @faustbrian @luciorubeens - please review this in the next few days. Be sure to explicitly select labels so I know what's going on.

dated:
The problem with this is that the shown information has no real value for the most part, since the current exchange rate is being used. Showing the historic exchange rate would require an API call for each shown transaction in the worst case, and a local storage lookup in the best. Iirc that is the reason why it was decided against this feature at the time.

j-a-m-l:
I know, but we have the same problem on the transaction page, which implements it, and the desktop wallet everywhere.

This would be helpful, although inaccurate, when seeing the latest transactions.

j-a-m-l:
@dated are you going to request the historic price on each block on #581 ?

If that's the case, I'll revamp this PR after that to work in a similar fashion (trying to not trigger lots of requests).

If not, what if we work together on that? We could start with the transaction page first.

dated:
Yea, but only the the block details page for now.

j-a-m-l:
OK, then I'll wait to update this PR after yours. Thanks

dated:
I guess local storage lookups aren't that hurtful. In the blocks table most of the time the blocks are from the same day, so it could be added on the tables as well.

dated:
Local storage seems to be precached from the browser upon visiting a domain, since the exchange rates are stored once per day the calls to Crypto Compare are kept at a minimum. I'll add the tooltips with the historic values to the tables later and might introduce a cap to the amount of stored rates, as to reduce bloat and limit the loading time upon first visit. Does that sound ok?

j-a-m-l:
I was writing something similar :wink: 

Yes, using the daily average would avoid lots of requests and use little local space. We could change the tooltips to reflect that the price is approximate.

That cap is a nice idea, although maybe it wouldn't be necessary, since `localStorage` could reach, depending on the browser, 2MB (mobile) to 10MB, which should be more than enough.
We could also catch any `Error` when trying to store new items, and then flush older ones, solely in that moment. As you prefer.

dated:
To be clear, the daily average is already used and stored :D

j-a-m-l:
Yes, I know :D

dated:
If you have too much on your plate I can add the same logic as in #581 to this PR @j-a-m-l, let me know.

ghost:
The [ci/circleci: test-node-10](https://circleci.com/gh/ArkEcosystem/explorer/783?utm_campaign=vcs-integration-link&amp;utm_medium=referral&amp;utm_source=github-build-link) job is failing as of [72a8ff2d658b55ac5a734eb73242733280ef042b](https://github.com/ArkEcosystem/explorer/commit/72a8ff2d658b55ac5a734eb73242733280ef042b). Please [review the logs for more information](https://circleci.com/gh/ArkEcosystem/explorer/783?utm_campaign=vcs-integration-link&amp;utm_medium=referral&amp;utm_source=github-build-link).

_Once you've pushed the fixes, the build will automatically re-run. Thanks!_

j-a-m-l:
@luciorubeens, in the last commit I've resolved the conflicts that this PR was having: basically they were caused by style/linting issues. Apart from that, I've included more linting suggestions.

j-a-m-l:
@ItsANameToo

ItsANameToo:
:mindblown: will look at this today

ItsANameToo (approved):

PR #681, opened 2019-07-22
chore(readme): add lead maintainer information

faustbrian:
Add information about how is the lead maintainer of the project/repository.


ItsANameToo:
need to check what's up with the pagination changes :sadup:

PR #685, opened 2019-07-25
fix: update webpack deps to fix broken state and disable e2e tests

ItsANameToo:
For some reason I failed to properly installed the latest dependencies in #677 , which broke `yarn dev`  on `develop` while tests ran fine. 
Updated the webpack dependencies to resolve these issues, but the e2e tests break as soon as we go above `webpack-dev-server: 3.2.0` (e.g. 3.2.1 and onwards) for an unknown reason. Disabled e2e for now until we move over to a different framework.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,11 +37,6 @@ jobs:
           when: on_success
           command: bash <(curl -s https://codecov.io/bash)
       - run:
-          name: e2e tests
-          command: yarn e2e
-          environment:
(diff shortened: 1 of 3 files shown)


faustbrian (approved):

faustbrian:
Marking this as draft as it requires the PHP 8 upgrade for the explorer servers. Doing that after the holidays.

faustbrian:
Just updated the server to PHP 8.0 so should be fine to merge.

PR #692, opened 2019-07-29
Merge develop into master

ItsANameToo:


ghost:
Your pull request doesn't follow our [contribution guidelines](https://docs.ark.io/guidebook/contribution-guidelines/). Please review and correct it.

Highjhacker:
@faustbrian Would it make sense to have the following to avoid repeating over and over the same constructor for all the `OrderByXScope`; 

```php
final class OrderByScope implements Scope
{
    public function __construct(
        protected string $direction = 'desc',
    ) {}
}

...

final class OrderByAmountScope extends OrderByScope
{
    public function apply(Builder $builder, Model $model)
    {
        $builder->orderBy('amount', $this->direction);
    }
}
```

Edit: https://github.com/ArkEcosystem/explorer.ark.io/pull/692/commits/637fb8f01fd591bdce4abb47f920d6a2a465946a

faustbrian:
You could consider to completely drop the order argument and simply have 2 sorting scopes for each field. 1 for ascending sorting and another for descending sorting, then it's only about calling the right class without any arguments.

Highjhacker:
Yeah I did considered that as first, but was a bit bothered by adding 30+ new scopes 🤔 I'll think about it

Edit: Thought about it since yesterday and I'm not convinced by that solution. It's just adding twice as much scope classes (so going to be around 36 just for the ordering), which is going to be annoying for things like autocompletion.

So, except if you or @ItsANameToo really want that instead of having an optional parameter, I'll stay with that.

faustbrian:
Those classes aren't complex to maintain so there shouldn't really be any overhead to them but not sure what @ItsANameToo thinks

ItsANameToo:
@Highjhacker failing tests

app/Http/Livewire/Concerns/BlocksOrdering.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    public string $blocksOrdering          = OrderingTypeEnum::HEIGHT;
+
+    public string $blocksOrderingDirection = OrderingDirectionEnum::DESC;
+
+    public function orderBlocksBy(string $value): void
+    {
+        $this->blocksOrdering = $value;
+
+        $this->blocksOrderingDirection = $this->blocksOrderingDirection === OrderingDirectionEnum::DESC ? OrderingDirectionEnum::ASC : OrderingDirectionEnum::DESC;
```
ItsANameToo: this will toggle every time you change a value, so you start with `timestamp:asc`, then click amount and it will sort by `amount:desc`, then click fee and it will sort `fee:asc`, etc so there's no default sort to start from (e.g. always asc then desc) which is confusing

ItsANameToo: this applies to all ordering classes

app/Models/Scopes/OrderByAmountAscScope.php
```diff
@@ -0,0 +1,17 @@
...
+
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Scope;
+
+final class OrderByAmountAscScope implements Scope
+{
+    public function apply(Builder $builder, Model $model)
+    {
+        $builder->orderBy('amount', 'asc');
```
ItsANameToo: doesn't seem right

image

resources/views/components/tables/desktop/delegates/active.blade.php
```diff
@@ -2,12 +2,12 @@
     <table>
         <thead>
             <tr>
-                <x-tables.headers.desktop.text name="general.delegates.rank" />
-                <x-tables.headers.desktop.address name="general.delegates.name" />
+                <x-tables.headers.desktop.text name="general.delegates.rank" wire:click="$emit('orderDelegatesBy', '{{ OrderingTypeEnum::RANK }}')" />
+                <x-tables.headers.desktop.address name="general.delegates.name" wire:click="$emit('orderDelegatesBy', '{{ OrderingTypeEnum::ADDRESS }}')" />
                 <x-tables.headers.desktop.status name="general.delegates.status">
```
ItsANameToo: missing status and productivity sorting

app/Models/Scopes/OrderByAmountAscScope.php
Highjhacker: Can't reproduce that one @ItsANameToo , it's on the transactions page ? Everything is properly ordered on my side 🤔

ItsANameToo: Homepage > view more > then sort amount asc and it shows that order for me

Highjhacker: Ok yeah indeed, will have a look. It have a hard time with `0` apparently

app/Http/Livewire/Concerns/TransactionsOrdering.php
```diff
@@ -27,9 +27,27 @@ trait TransactionsOrdering
...
+        if ($value === $this->transactionsOrdering) {
+            if ($this->transactionsOrderingDirection === OrderingDirectionEnum::DESC) {
+                return 'chevron-down';
+            }
+
+            return 'chevron-up';
+        }
 
-        $this->transactionsOrderingDirection = $this->transactionsOrderingDirection === OrderingDirectionEnum::DESC ? OrderingDirectionEnum::ASC : OrderingDirectionEnum::DESC;
+        return 'chevron-down';
```
ItsANameToo: this should be a different icon, otherwise it looks like a column is sorted while it isn't

app/Http/Livewire/Concerns/TransactionsOrdering.php
```diff
@@ -27,9 +27,27 @@ trait TransactionsOrdering
...
+            $this->transactionsOrderingDirection =
+                $this->transactionsOrderingDirection === OrderingDirectionEnum::DESC ? OrderingDirectionEnum::ASC : OrderingDirectionEnum::DESC;
+        }
+
         $this->transactionsOrdering = $value;
+    }
+
+    public function renderDirectionIcon(string $value)
+    {
+        $value = substr($value, strrpos($value, '.') + 1);
```
ItsANameToo: what's this for?

Highjhacker:
Will go back into it a bit later today, got some issue that I need to figure out first. Issue being the sorting icon for columns. I'd like to make it visible on one column at the time when that column header was clicked, but haven't figured out a good way to do it for now, for now I'm only able to have the `chevron` icons displayed all the time.

Highjhacker:
Still a couple of things to do, but as for the ordering itself, I think all tables are good.

Remaining points : 

- Delegate ordering is quite slow, even if it's only applying it on the active delegates of the page.
- Should we apply ordering to the `Monitor` table ? Didn't went for that at first cause it seems to be a similar situation to the homepage where we don't allow it.
- Check to clean up 
- Tests
- ~~Icon when no ordering is applied~~ Should display `chevron-down` on hover when no ordering is applied on a column according to styleguide.

app/Http/Livewire/Concerns/TransactionsOrdering.php
Highjhacker: If I remember correctly, it was to have "unique" directions based on the column, so clicking timestamp column doesn't change the direction of another column. And the whole `substr` thingy is because I get the value like `"general.transaction.timestamp"` so need to parse it.

ItsANameToo:
will get to this later today

alfonsobries:
When I sort the table it shows the skeleton with a weird jump:


https://user-images.githubusercontent.com/17262776/120507198-35c6f080-c38c-11eb-9fff-0b48f879ee82.mov

alfonsobries:
The sort icon changes the column width

https://user-images.githubusercontent.com/17262776/120507432-6d359d00-c38c-11eb-9db3-294e201f5f53.mov

alfonsobries:
This is more like a design change but there is no clue that I can sort the columns unless I hover the header 
image

alfonsobries:
maybe is something related to the playbook database but after I sorted the blocks table by timestamp it showing me some cells without the view icons but there a link there (if I move the cursor over the cell It has a link to the transaction). 
image

alfonsobries:
Similar issue in the transaction table (notice the timelock icons)
image

alfonsobries:
Related to design also but think we need an indication that the table is sorted by a specific column.

PR #722, opened 2019-09-03
chore: improve gh pages script

ItsANameToo:
Use the `base` parameter when building for gh pages instead of adjusting the webpack config to handle it.


ItsANameToo:
Nice commits @ItsANameToo, please rebase

PR #724, opened 2019-09-03
feat: replace delegate status icons

### LINKED ISSUE #693: Accessibility: Status images in the delegate table are not keyboard accessible
**Expected Behavior**

A screen reader, or even a sighted user who is only using a keyboard, should be able to understand the "status" column in the main Delegate table here:

https://dexplorer.ark.io/delegate-monitor

**Current Behavior**

The current popover for Status seems to be an mouse-only 'on-hover' event, which is not accessible for keyboard-only users, including the blind and those who are unable to use a mouse. The image itself does not contain any alt text, and a screen reader simply says "image" when navigated to, and the popover is not triggered.

**Possible Solution**

Please either add some alt text that mimics what's in the popover, or trigger the popover on-focus on that image in the table so the screen reader can announce it. Example, include's aria-role and aria-describedBy attributes:
https://heydonworks.com/practical_aria_examples/#input-tooltip

### LINKED ISSUE #695: Accessibility: Status images in the delegate table are not suitable for users with impaired acuity
**Expected Behaviour**

Users should be **easily** able to tell the difference between the three statuses.

**Current Behaviour**

It is not possible for certain colourblind people with additional acuity problems to tell the difference. Try crossing your eyes and looking at the explorer, imagining there is no colour.

**Possible Solution**

1. Use different larger shapes. A full sized green square, red circle or yellow triangle for the respective statuses would provide more definition to the icons rather than dots and dashes. Different shapes would be more easily differentiated for users with acuity problems.

Alternatively:

2. Just use ✅, ❌ and ⚠️ unicode emoji which will render natively and already convey enough visual information to be understandable without trying to work out "what do two dots, a dot and a dash or two dashes mean" as they are already intuitive in context. Bonus points for the fact that they work natively with screen readers re: #693. No need to reinvent the wheel.

dated:
- [x] replaces the delegate status icons (closes #695)
- [x] improves accessibility of the delegate monitor (closes #693)

image

## What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [x] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## Does this PR release a new version?

- [ ] Yes
- [x] No

If yes, please describe the impact and migration path for existing applications:

**The PR fulfills these requirements:**

- [x] It's submitted to the `develop` branch, _not_ the `master` branch
- [x] All tests are passing
- [ ] New/updated tests are included

If adding a **new feature**, the PR's description includes:

- [x] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)


alessiodf:
Are these the same design that @ItsANameToo showed us a few weeks ago?

ItsANameToo:
Yes @alessiodf

alessiodf:
This may be out of scope but is it not worthwhile having a separate icon (like, for example, a blue circle) to denote the _never forged_ delegates rather than showing the not forging symbol?

ItsANameToo (dismissed):

ghost:
A contributor has approved this PR. A maintainer will merge this PR shortly. If it shouldn't be merged yet, please leave a comment saying so and we'll wait.

ItsANameToo (approved):

ghost:
A contributor has approved this PR. A maintainer will merge this PR shortly. If it shouldn't be merged yet, please leave a comment saying so and we'll wait.

ItsANameToo:
conflict :trollface:

ghost:
Your pull request has been merged and marked as tier 4. _It will earn you $20 USD._

app/Http/Livewire/PriceStats.php
```diff
@@ -0,0 +1,79 @@
...
+        $priceFullRange = CryptoCompare::historicalHourly(Network::currency(), Settings::currency(), 24);
+
+        $initialPrice = $priceFullRange->first();
+
+        return $initialPrice === 0 ? 0 : ($priceFullRange->last() / $initialPrice) - 1;
+    }
+
+    private function getHistorical(): Collection
+    {
+        if ($this->placeholder) {
```
alfonsobries, line 53: These values are for creating the chart that is used as a placeholder:

image

samharperpittam:
@alfonsobries 

Have some questions on this PR.

1. With the `.env` set to `EXPLORER_NETWORK=development`, the Explorer is then using our devnet. To which, as DARK tokens do not have any value, the price stat should probably use the placeholder? 

2. With the `.env` set to `EXPLORER_NETWORK=production`, the Price stat is showing a price of $7.38 which is incorrect. Is it pulling in the price from cryptocompare now? 

3. When viewing stats which are negative, the percentage value doesn't appear to be aligned with the price value as shown in the image below (probably only a few pixels out). 
image

4. It appears to me that the stats are showing the refresh animation every 8 seconds rather than every 60?

alfonsobries:
@samharperpittam 

1. The placeholder is shown according to the value in the `explorer.development.canBeExchanged`(`config/explorer.php`) setting. 

If `EXPLORER_NETWORK_CAN_BE_EXCHANGED` is set to `development`. The value of that setting is `false` by default so you should see the placeholder. If you are seeing something different you probably have the value of`EXPLORER_NETWORK_CAN_BE_EXCHANGED` set to `true`, simply remove that setting from the `env` file (or set it to `false`).

image


2. that value comes from cryptocompare maybe you need to remove the cache: `php artisan cache:clear` (I see 1.45)

3. and 4. working on it

alfonsobries:
@samharperpittam just fixed points 3 and 4, so just waiting for your comment on the rest of the feedback

samharperpittam:
Thanks for the info @alfonsobries  - Much appreciated!

Everything is working as expected now.

My only other comment and it's pretty minor, is that when compared with the design, the gaps between each component in the grid feels a bit large.

**Design**
image

**Implementation**
image

If we can bring that a little closer to the designs then I'm happy with this 👍

alfonsobries:
@samharperpittam ✅

resources/lang/en/general.php
```diff
@@ -3,26 +3,28 @@
...
-    'wallet_not_found' => '<span class="bg-theme-warning-100">:0</span> has no balance. <br/> <span class="text-base font-normal">Return to this page after the address has received a transaction.</span>',
-    'more_details'     => 'For more :transactionType details',
-    'learn_more'       => 'Learn more',
-    'confirmations'    => ':count confirmation|:count confirmations',
+    'address'                 => 'Address',
+    'beta_uppercase'          => 'BETA',
+    'optional'                => 'Optional',
+    'or'                      => 'or',
+    'explorer'                => 'Explorer',
+    'filler'                  => '',
```
ItsANameToo: what's this for? 🤔

resources/views/livewire/price-stats.blade.php
```diff
@@ -0,0 +1,163 @@
...
+        </div>
+    @endunless
+
+
+    <div class="hidden flex-grow justify-end lg:flex" >
+        <div
+            class="ml-6 h-10"
+            style="width: 120px;"
+            @toggle-dark-mode.window="toggleDarkMode"
+            x-data="{
```
ItsANameToo: can this be moved to a js file? would make it easier to maintain than having it in the php file

resources/views/livewire/network-status-block.blade.php
```diff
@@ -1,14 +1,31 @@
...
-            <div class="hidden sm:block">@lang('general.market_cap'): <x-currency :currency="Network::currency()">{{ $marketCap }}</x-currency></div>
-
-            <div class="md:hidden">
-                <livewire:price-ticker />
+<div class="overflow-auto bg-theme-secondary-100 dark:bg-black ">
+    <div class="py-10 content-container-full-width">
+        <div class="px-8 md:w-full">
+            <div
+                class="flex grid-cols-2 gap-3 w-full md:grid xl:space-x-3 xl:gap-0 xl:flex"
+                wire:poll.60s
```
ItsANameToo: can this be changed to poll every 8 seconds (block time) so height/supply don't go out of sync for a minute while new blocks are being generated? Preferably the polling timeframe here should be taken from a configuration file so it's easy to adjust if needed (e.g. in case of an ark fork having a different block time). Not sure how well livewire works with that

alfonsobries: @ItsANameToo take it the from `Network::blockTime()`

alfonsobries:
@ItsANameToo should be ready now. I wasn't able to replicate the issue you had with the chart out of bonds but refactored the chart to have a fixed size and ensured to set the min/max value of the chart, can you test it again?

PR #732, opened 2019-09-12
fix: qr-code

ItsANameToo:


faustbrian (approved):

app/Http/Livewire/Stats/InsightAllTimeTransactions.php
```diff
@@ -0,0 +1,89 @@
...
+use Illuminate\View\View;
+use Livewire\Component;
+
+final class InsightAllTimeTransactions extends Component
+{
+    use AvailablePeriods;
+
+    public string $period = 'week';
+
+    private string $chartColor = 'grey';
```
ItsANameToo: 🤔 can this be changed to be handled in css so we can use the tailwind classes for it? or at the minimum see if we can use the proper hex codes to it's in line with the grey and yellow colors we use normally

resources/views/components/stats/insight.blade.php
```diff
@@ -0,0 +1,67 @@
+@props([
+    'title',
+    'value',
+    'title2',
+    'value2',
+    'title3' => null,
+    'value3' => null,
```
ItsANameToo: these props don't tell me what they are used for, can you give them a more descriptive name?

resources/views/components/stats/insight.blade.php
```diff
@@ -0,0 +1,67 @@
...
+    'chart' => null,
+    'chartColor' => 'grey',
```
ItsANameToo: I don't see this used in the component? 🤔

resources/views/components/stats/insight.blade.php
```diff
@@ -0,0 +1,67 @@
...
+                    <p class="mt-2 text-base font-semibold text-theme-secondary-700 dark:text-theme-secondary-200">{{ $value3 }}</p>
+                </div>
+            @endif
+        </div>
+    </div>
+
+</x-general.card>
+
+@push('scripts')
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/dayjs/1.9.1/dayjs.min.js"></script>
```
ItsANameToo: what is dayjs used for?

app/Http/Livewire/Stats/InsightAllTimeTransactions.php
```diff
@@ -0,0 +1,89 @@
...
+            return Cache::remember($cacheKey, (int) $this->refreshInterval, fn () => (string) Transaction::count());
+        }
+
+        [$from, $to] = $this->getRangeFromPeriod($period);
+        $cacheKey .= ".{$from }.{$to}";
+
+        return Cache::remember($cacheKey, (int) $this->refreshInterval, fn () => Transaction::query()
+                ->select(DB::raw("to_char(to_timestamp(timestamp), 'yyyy-mm-dd') as period, COUNT('id') as transactions"))
+                ->whereRaw("to_char(to_timestamp(timestamp), 'yyyy-mm-dd') > ?", [$from])
+                ->whereRaw("to_char(to_timestamp(timestamp), 'yyyy-mm-dd') <= ?", [$to])
```
ItsANameToo: is `to` needed? doesn't it suffice fetch all the transactions that have a timestamp larger than the period we're looking for? e.g. week means larger than 1 week ago until now

ItsANameToo: note that `timestamp` for ark is different than an epoch timestamp. Instead it starts at `0` for the first transaction as it uses its own epoch. Basically you need to subtract the epoch data from all current timestamps to get the timestamp that is used in the column.

basically you need: `now - ark epoch = timestamp` where ark epoch is `1490101200`

app/Http/Livewire/Stats/InsightCurrentAverageFee.php
```diff
@@ -0,0 +1,89 @@
...
+            $cacheKey .= ".{$from }.{$to}";
+        }
+
+        return Cache::remember($cacheKey, (int) $this->refreshInterval, fn () => Transaction::query()
+                ->select(DB::raw("to_char(to_timestamp(timestamp), 'yyyy-mm-dd') as period, AVG(fee / 1e8) as average, MIN(fee / 1e8) as minimum, MAX(fee / 1e8) as maximum"))
+                ->when($period === 'current', function ($query): void {
+                    $query->whereRaw("to_char(to_timestamp(timestamp), 'yyyy-mm-dd') = ?", [Carbon::now()->toDateString()]);
+                }, function ($query) use ($from, $to): void {
+                    $query
+                        ->whereRaw("to_char(to_timestamp(timestamp), 'yyyy-mm-dd') > ?", [$from])
```
ItsANameToo: same epoch story here

resources/views/components/stats/insight.blade.php
leMaur: Not for now. I have to find a way to pass the color info inside the chart. Still in my TODO list. 😇

resources/views/components/stats/insight.blade.php
leMaur: Whoops, forgot to remove.

alfonsobries:
Yellow chart over text (my browser width was 1313px)
image

alfonsobries:
Unaligned labels: (apply to the three charts)

Screen Shot 2021-06-02 at 10 53 30

alfonsobries:
Chart is not being resized when browser is resized 
image

alfonsobries:
Medium screen (768px) charts 

Design:
image

Implementation:
image

alfonsobries:
On sm screen (640px)

image


** In general some similar issues with alignments on different screen sizes

app/Http/Livewire/Concerns/AvailablePeriods.php
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        return Carbon::createFromTimestamp((int) Carbon::now()->timestamp - $this->getArkEpoch())
+            ->sub($this->subtractFromPeriod($period))
+            ->toDateString();
+    }
+
+    private function getArkEpoch(): int
+    {
+        return 1490101200;
```
alfonsobries: Move this to a constant  or to settings file, I would ask to @ItsANameToo maybe that variable even already exists

app/Http/Livewire/Stats/InsightAllTimeFeesCollected.php
```diff
@@ -0,0 +1,104 @@
...
+            'chartValues'               => $this->chartValues($this->period),
+            'chartTheme'                => $this->chartTheme(),
+            'options'                   => $this->availablePeriods(),
+            'refreshInterval'           => $this->refreshInterval,
+        ]);
+    }
+
+    private function allTimeFeesCollected(): string
+    {
+        $value = $this->transactionsPerPeriod('all-time');
```
alfonsobries: I would move that period to a constant

app/Http/Livewire/Stats/InsightAllTimeTransactions.php
```diff
@@ -0,0 +1,97 @@
...
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Facades\DB;
+use Illuminate\View\View;
+use Livewire\Component;
+
+final class InsightAllTimeTransactions extends Component
+{
+    use AvailablePeriods;
+
+    public string $period = 'week';
```
alfonsobries: Move period to a constant?

resources/js/chart-theme.js
```diff
@@ -0,0 +1,176 @@
...
+    return {
+        grey: colors.grey,
+        black: colors.black,
+        yellow: colors.yellow,
+        green: colors.green,
+        red: colors.red,
+    };
+}
+
+export function hexToRgb(hex) {
```
alfonsobries: Since we are hardcoding those colors may make sense to add them as rgb initially so we can get rid of this conversion?

leMaur:
> Unaligned labels: (apply to the three charts)
> 
> Screen Shot 2021-06-02 at 10 53 30

@alfonsobries 
I've added `-ml-1` because on Firefox was not aligned but I forget to test it on other browsers.

Fun fact https://bugzilla.mozilla.org/show_bug.cgi?id=1582545 an old bug (2 years). 
It seems that Firefox add 4px inline padding [source](https://searchfox.org/mozilla-central/source/layout/style/res/forms.css#310).

Adding `-moz-transform: translateX(-4px);` seems to solve the problem, but it sounds dirty.
Do you have an idea on how to handle it better?

app/Http/Livewire/Concerns/AvailablePeriods.php
leMaur: @alfonsobries 👍🏽  Not a variable but a method returning a `Carbon` instance `Network::epoch()`

app/Http/Livewire/Stats/InsightAllTimeTransactions.php
leMaur: @alfonsobries In the `mount` method it already apply a default period. I just added it to have a possibility to override it in the future if necessary.
In my opinion I can initialize the variable with an empty string. What do you think?

resources/js/chart-theme.js
leMaur: @alfonsobries I've added because I wanted to use the values from tailwind config colors. But I doubt if it's good or not. What's your thoughts? (See [Mici's comment](https://github.com/ArkEcosystem/explorer.ark.io/pull/732#discussion_r641599973))

app/Http/Livewire/Concerns/AvailablePeriods.php
```diff
@@ -0,0 +1,57 @@
...
+
+    private function subtractFromPeriod(string $period): string
+    {
+        return collect([
+            'day'     => '1 day',
+            'week'    => '1 week',
+            'month'   => '1 month',
+            'quarter' => '1 quarter',
+            'year'    => '1 year',
+            'all'     => '100 years',
```
ItsANameToo: what is this for

app/Http/Livewire/Stats/InsightAllTimeFeesCollected.php
```diff
@@ -0,0 +1,106 @@
...
+use Illuminate\View\View;
+use Livewire\Component;
+
+final class InsightAllTimeFeesCollected extends Component
+{
+    use AvailablePeriods;
+
+    private const PERIOD_ALL_TIME = 'all-time';
+
+    private const CHART_COLOR = 'yellow';
```
ItsANameToo: 🤔

app/Http/Livewire/Stats/InsightAllTimeFeesCollected.php
```diff
@@ -0,0 +1,106 @@
...
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Facades\DB;
+use Illuminate\View\View;
+use Livewire\Component;
+
+final class InsightAllTimeFeesCollected extends Component
+{
+    use AvailablePeriods;
+
+    private const PERIOD_ALL_TIME = 'all-time';
```
ItsANameToo: can periods be moved to an enum so they can't result in typos? also noticed that here it's `all-time` while in another file it's `all` so that can be confusing

resources/js/chart-theme.js
ItsANameToo: I had hoped we could reuse it in some way, but if it takes this much effort to make use of the hex values then we are better off with the hardcoded rgb ones :sadup:

config/explorer.php
```diff
@@ -36,6 +36,15 @@
         ],
     ],
 
+    'node-fees' => [
+        'production' => [
+            'endpoint' => 'https://wallets.ark.io/api/node/fees',
```
ItsANameToo: you can query the database for this directly. The endpoint was just to illustrate what kind of values core reports

app/Enums/StatsTransactionTypes.php
```diff
@@ -0,0 +1,34 @@
...
+
+declare(strict_types=1);
+
+namespace App\Enums;
+
+final class StatsTransactionTypes
+{
+    public const ALL = 'all';
+
+    public const TRANSFER = 'transfer';
```
ItsANameToo: I assume these are used for the api endpoint?

app/Http/Livewire/Stats/InsightAllTimeFeesCollected.php
ItsANameToo: still " 🤔 "

config/explorer.php
leMaur: @ItsANameToo I've modified the `FeeCache.php`. Now it accepts an optional transaction type.
In this way, the data are kept in cache for fast access.

app/Http/Livewire/Stats/InsightAllTimeFeesCollected.php
leMaur: I've used as a configuration key instead of scrolling until [`chartTheme`](https://github.com/ArkEcosystem/explorer.ark.io/pull/732/files/3cbdacb65f6a0a20203f7d313070fc8eafca3dff#diff-01f15afc63c6972086167d958466637d782b5c28e55c8fa7aeaca3852216dcdaR78).
Or do you mean something else?

app/Console/Commands/CacheFees.php
```diff
@@ -43,5 +43,13 @@ public function handle(FeeCache $cache)
 
             $cache->setMaximum($period, MaximumAggregateFactory::make($period)->aggregate());
         }
+
+        foreach (['transfer', 'secondSignature',  'delegateRegistration',  'vote',  'multiSignature',  'ipfs',  'multiPayment',  'delegateResignation',  'timelock',  'timelockClaim',  'timelockRefund',  'magistrate'] as $type) {
```
ItsANameToo: check how this is shown in other locations (e.g. homepage has a dropdown for types, so does search) and make it all use a generic location to fetch transactions from to make it more maintainable and not scattered throughout the codebase

app/Enums/StatsTransactionTypes.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Enums;
+
+final class StatsTransactionTypes
```
ItsANameToo: what's this file used for exactly? I see it being mapped to something in the `AvailableTransactionType` component, but why use these instead of the existing `CoreTransactionTypeEnum` for example?

app/Http/Livewire/Stats/InsightAllTimeFeesCollected.php
```diff
@@ -0,0 +1,77 @@
...
+    private function chartValues(string $period): Collection
+    {
+        return $this->transactionsPerPeriod($period);
+    }
+
+    private function chartTheme(): Collection
+    {
+        $mode = Settings::usesDarkTheme() ? 'dark' : 'light';
+
+        return collect(['name' => self::CHART_COLOR, 'mode' => $mode]);
```
ItsANameToo: this will require a page refresh when switching between dark/light mode as the colors won't be applied immediately upon toggling the setting. Needs to be reconsidered to make it work upon changing the setting

app/Http/Livewire/Stats/InsightAllTimeTransactions.php
```diff
@@ -0,0 +1,76 @@
...
+use App\Enums\StatsPeriods;
+use App\Http\Livewire\Concerns\AvailablePeriods;
+use App\Services\Cache\TransactionCache;
+use App\Services\NumberFormatter;
+use App\Services\Settings;
+use Illuminate\Support\Collection;
+use Illuminate\View\View;
+use Livewire\Component;
+
+final class InsightAllTimeTransactions extends Component
```
ItsANameToo, line 15: this file looks very similar to the `InsightAllTimeFeesCollected.php` one, would it be possible to make it an abstract class and simply override the `transactionsPerPeriod` in classes that implement it to make it point to the correct cache to pull the data from? 🤔

ItsANameToo: I suppose it needs a formatter too but that could be moved to a separate method too so it can be overridden

app/Http/Livewire/Stats/InsightCurrentAverageFee.php
```diff
@@ -0,0 +1,76 @@
...
+
+use App\Facades\Network;
+use App\Http\Livewire\Concerns\AvailableTransactionType;
+use App\Services\Cache\FeeCache;
+use App\Services\NumberFormatter;
+use Illuminate\Support\Collection;
+use Illuminate\View\View;
+use Livewire\Component;
+
+    final class InsightCurrentAverageFee extends Component
```
ItsANameToo: indent

app/Http/Livewire/Stats/InsightAllTimeTransactions.php
ItsANameToo: maybe it's only 2 `Insight` classes that are similar, so might not be too relevant in the end but something to consider :oleglala:

resources/js/chart-theme.js
ItsANameToo: can you change the hex values to rgb directly? if we need these methods to get that handled it's much easier to do the conversion manually and add a comment to the colors instead of what they resemble in tailwind (e.g. `borderColor: <rgb value>, // gray-900`) for easy lookup

ItsANameToo: additional question, how are colors currently used for the homepage chart (the small one at the top) as that uses green/red colors too 🤔

app/Services/Transactions/Aggregates/Fees/Maximum/AllAggregate.php
```diff
@@ -0,0 +1,34 @@
...
+    private ?string $type;
+
+    public function setType(?string $type = null): self
+    {
+        $this->type = $type;
+
+        return $this;
+    }
+
+    public function aggregate(): float
```
ItsANameToo: will take a look why the `all` aggregates always result in 0 for me

app/Http/Livewire/Stats/InsightCurrentAverageFee.php
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        private function maxFee(string $transactionType): string
+        {
+            $value = $this->getFeesAggregatesPerPeriod($transactionType);
+
+            return NumberFormatter::currency(data_get($value, 'max', 0), Network::currency());
+        }
+
+        private function getFeesAggregatesPerPeriod(string $transactionType): Collection
```
ItsANameToo: we're not using a period for min/avg/max fees so the name is a bit confusing

app/Console/Commands/CacheFees.php
leMaur: @ItsANameToo the dropdown in the homepage is [hard-coded](https://github.com/ArkEcosystem/explorer.ark.io/blob/050a56ac1495d0aacb6ba2c3ffd40711b181b610/resources/views/components/transaction-table-filter.blade.php#L8).
Question: the list contains "magistrate" and (of course) it's not in `CoreTransactionTypeEnum`, this is why I've created `StatsTransactionTypes`. Do I have to remove "magistrate" from the list (so showing only Core transactions)?

app/Console/Commands/CacheFees.php
```diff
@@ -43,5 +43,13 @@ public function handle(FeeCache $cache)
...
+            $cache->setMinimum($type, MinimumAggregateFactory::make('all', $type)->aggregate());
+
+            $cache->setAverage($type, AverageAggregateFactory::make('all', $type)->aggregate());
+
+            $cache->setMaximum($type, MaximumAggregateFactory::make('all', $type)->aggregate());
```
ItsANameToo: for min/avg/max fees we don't want to go all the way back. All we want to know for these values is the values over the last 20 transactions to mimic how the `core` endpoint works (see https://dwallets.ark.io/api/node/fees). For example for votes you would have the following query to get the last 20 and fetch the average for it (`select avg(fee) from (select id, fee from transactions where type = 3 and type_group = 1 order by "timestamp" desc limit 20) as fees;`):

image

this is in line with what the core endpoint reports for `vote` `avg`:
```
"vote": {
        "avg": "19434771",
        "max": "46830489",
        "min": "1000000",
        "sum": "388695412"
      },
```

ItsANameToo: same will apply to all avg/min/max calculations for fee types. Currently the way the eloquent builder is set up it doesn't report these kind of values to me

app/Console/Commands/CacheFees.php
ItsANameToo: you can keep `magistrate`, as that's basically everything with `type_group` set to `2` (we have `1` for core transactions, `2` for magistrate ones which are currently limited to entity update/registration/resignation, see https://dwallets.ark.io/api/node/fees). You can create a query to target the type_group and group all the magistrate transactions together for the stats

ItsANameToo: regarding the hard-coded list, can you move that out of the component to a generic place where it can be reused? will make it easier to update etc :oleglala:

app/Enums/StatsTransactionTypes.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    public const TRANSFER = 'transfer';
+
+    public const SECOND_SIG = 'secondSignature';
```
ItsANameToo: can you make the enums full words? e.g. `SECOND_SIGNATURE`. Applies to a couple of them

resources/js/chart-theme.js
leMaur: @ItsANameToo Actually are hard-coded inside the [PriceChart](https://github.com/ArkEcosystem/explorer.ark.io/blob/develop/resources/js/price-chart.js#L39) component. As soon as we merge this PR, I can prepare a task to change the homepage chart with the new component. The colors used are the same since I've picked from it and double-checked with the design.

May we export the chart component to the UI library (with its proper documentation on how to use it)? Since it's a reusable component. 
What do you think?

app/Console/Commands/CacheFees.php
leMaur: @ItsANameToo I've reused the class created by Alex on PR#865 https://github.com/ArkEcosystem/explorer.ark.io/pull/865/files#diff-cac35ee1cf2926a57e64a967b2d2c3f8a6f94ad174132a379017ca23f464a584. 
But it doesn't contain **delegateResignation** and I don't know if it was done on purpose.

resources/js/chart-theme.js
ItsANameToo: can move it to a UI component after this gets merged and it needs to get reused for the homepage one :oleglala:

app/Console/Commands/CacheFees.php
ItsANameToo: @leMaur good catch, must have been an oversight as it was also not in the search list but was part of the transaction list here https://github.com/ArkEcosystem/explorer.ark.io/pull/865/files#diff-0813a7b84ce26ec1502030e923bab886aa1fa8ff3dce9ace038a765adbca261eL18

can you re-add that please

alexbarnsley: hmm good spot. i copied this list from somewhere else 🤔

alexbarnsley: https://github.com/ArkEcosystem/explorer.ark.io/pull/865/files#diff-08956e04d3fca1cf849bcef92f0c3e8711b82b628d7f3b234aa51c273599a955L37

from the search module

ItsANameToo:
image
in addition to the discussion on slack about the chart, can you hide the `timelock claim` and `timelock refund` options in the dropdown here? as those will always have `0` for fees. You can keep the `timelock` one and this change will be limited to the stats page as the other dropdowns will still require the ability to sort by htlc claim/refunds, only for the stats page it should be hidden

app/Services/Transactions/Aggregates/Fees/MinimumAggregateFactory.php
```diff
@@ -38,6 +40,14 @@ public static function make(string $period)
             return new YearAggregate();
         }
 
+        if (Str::of($period)->contains('last')) {
```
ItsANameToo, line 43: when is this used?

composer.json
```diff
@@ -42,6 +42,8 @@
         "furqansiddiqui/bip39-mnemonic-php": "^0.1.3",
         "konceiver/phpunit-speedtrap": "^1.0",
         "laravel/dusk": "^6.6",
+        "spatie/laravel-ray": "^1.23",
+        "spatie/ray": "^1.27",
```
ItsANameToo: can you remove this as general dep

app/Services/Transactions/Aggregates/Fees/MinimumAggregateFactory.php
leMaur: @ItsANameToo It's used in [CacheFee](https://github.com/ArkEcosystem/explorer.ark.io/pull/732/files#diff-391a4a5d3f8187e393e2b60fab1802d7bfa9f82d6a2acdee658206c7910d1b52R57-R59) to get only the latest 20 records

ItsANameToo: ah alright, thanks. Was looking for a `last` entry in the dropdown but that makes sense 😂

PR #733, opened 2019-09-12
fix: use correct process.env var in migration

ItsANameToo:


tests/AvailablePeriodsLivewireTest.php
```diff
@@ -0,0 +1,39 @@
+<?php
```
ItsANameToo: what's this

ItsANameToo:
@leMaur conflicts

ItsANameToo:
image

also the dropdown ends up behind the tab component

composer.json
```diff
@@ -42,6 +42,8 @@
         "furqansiddiqui/bip39-mnemonic-php": "^0.1.3",
         "konceiver/phpunit-speedtrap": "^1.0",
         "laravel/dusk": "^6.6",
+        "spatie/laravel-ray": "^1.23",
+        "spatie/ray": "^1.27",
```
ItsANameToo: booo mau

resources/views/components/stats/periods-selector.blade.php
```diff
@@ -0,0 +1,81 @@
...
+<div {{ $attributes->only('class') }}>
+
+    {{--mobile: input select--}}
+    <label class="relative sm:hidden">
+        <span class="flex absolute top-0 right-0 bottom-0 items-center transform translate-x-2 translate-y-px">
+            <x-ark-icon name="chevron-down" size="2xs" />
+        </span>
+        <select
+            style="-moz-transform: translateX(-4px);"
+            {{-- https://searchfox.org/mozilla-central/source/layout/style/res/forms.css#310 --}}
```
ItsANameToo: can you add a comment such as `translate is needed in firefox to avoid spacing issue`? then the links can stay too but the code itself has some additional information on what the fix is for

app/Http/Livewire/Stats/Chart.php
```diff
@@ -0,0 +1,137 @@
...
+
+    private function mainValuePercentage(): float
+    {
+        return abs($this->getDiffFromHistoricalHourly()) * 100;
+    }
+
+    private function mainValueVariation(): string
+    {
+        return $this->getDiffFromHistoricalHourly() < 0 ? 'down' : 'up';
+    }
```
ItsANameToo: can this rely on the same data as the homepage? so they can't go out of sync  like they do now (currently it can be indexed on the homepage while stats will show 0% for example)

PR #773, opened 2019-11-01
fix: use typeGroups in addition to type

### LINKED ISSUE #759: Refactor: use TypeGroup when checking tx types
With 2.6, `TypeGroups` are used in addition to the transaction types. Whenever we check for a specific transaction `type`, we should also assert that it is the correct `TypeGroup` so we don't end up with unexpected results. This is not done throughout all places where we check transaction types yet and needs to be improved.

ItsANameToo:
Resolves #759

### DIFF
--- a/src/components/SelectionType.vue
+++ b/src/components/SelectionType.vue
@@ -61,7 +61,7 @@
 
 <script lang="ts">
 import { Component, Prop, Vue } from "vue-property-decorator";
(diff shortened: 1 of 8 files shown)


dated:
:hide:

ItsANameToo:
what?

ItsANameToo:
@alfonsobries some conflicts :hide:

alfonsobries:
@ItsANameToo ✅

PR #774, opened 2019-11-01
fix: add timelock status to detail screen and update multipayment table

ItsANameToo:
Adds a timelock status entry to the transaction detail screen; it will link to claimed / refunded timelocks or show `open` for timelocks that had no interaction yet.

Also adds a background to the multipayment table to be in line with the other tables in the explorer


Highjhacker:
Ran the tests a bunch of times and no more fails, looks good to me :lala:

PR #777, opened 2019-11-21
feat: bignumber arithmetic

### LINKED ISSUE #705: BigNumber arithmetic
In some places the explorer performs calculations or arithmetic comparisons on the balances of wallets - like on the top wallets page and the currency amount tooltips.

**Current Behavior**

Since as of recently big numbers are returned as strings from the API Vue throws some unexpected type warnings.

**Possible Solution**

Implement a wrapper for big numbers to ensure calculations on those values deliver the desired result and to get rid of the type warnings.

ItsANameToo:
Resolves #705 

Will do another pass to check if other parts require it too although most calculations are isolated to the mixins.


ghost:
Your pull request doesn't have a test case, which is a requirement for it to be merged. Please provide it and one of the developers will review it before merging.

PR #781, opened 2019-11-28
fix: build sum over payments to same address

### LINKED ISSUE #780: Explorer amount for multi payment incorrect if address has more than one tx in single multipament transaction
**Expected Behavior**

Multipayment should sum up all transactions for the same address (in a single multipayment tx) so that explorer is reflective of the amount the wallet actually received, not just the first amount it finds. 

**Current Behavior**

The explorer only searches for the first instance of the wallet in the multipayment that it can find. Therefore if an address has more than one transaction in single multipayment it will not be reflective in the explorer as they will not be summed up. 

**Possible Solution**

Sum up all relevant transactions for an address in a single multi-payment and use that as the amount in explorer

**Steps to Reproduce (for bugs)**

1) Post multipayment with amounts sent to the same address 
2) view Explorer to see mismatch. Example is below. 

https://dexplorer.ark.io/transaction/a3e6998881cdca50d175bd4dd5613f4ac85beaa9f39d7048691130fbdcc896cf

**Your Environment**
- Version used: Core 2.6
- Operating System and version: Ubuntu 18.04

dated:
## Summary

Fixes #780.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


ItsANameToo:
where did codecov go 🤔

dated:
> where did codecov go 🤔

I don't know all my repos are outdated on codecov 😭

ItsANameToo:
One issue I noticed: it doesn't color the amount properly when the sender is also part of the multipayment. For an example: see `DMzBk3g7ThVQPYmpYDTHBHiqYuTtZ9WdM3` and `a3e6998881cdca50d175bd4dd5613f4ac85beaa9f39d7048691130fbdcc896cf`. The wallet will show `18,4204893` as outgoing amount, while the full multipayment is actually `19,7999998` in total of which (4,9 + 4,31024465 + 4,9 + 4,31024465 = 18,4204893) is for the sender itself 🤔 So it should actually show `19,8 - 18,4` as outgoing amount in that specific case. Right now the explorer assumes the multipayment is outgoing when it's the sender, while the calculation assumes the amount is based on the receiver only so there's a mismatch

dated:
Good catch, I'll adjust that.

ItsANameToo:
Will have to think a bit on how to properly show the multipayment transactions when they are both incoming and outgoing at the same time, which complicates things a bit (e.g. if you click on `received` it will show them as outgoing amounts, things like that)

dated:
Oh, right.

ItsANameToo:
What do you think of having a tooltip when you hover over the amount that shows the full amount of the multipayment (only if the amount shown is different from the total)? I'm sure people will be confused when they send a multipayment with 10 ark and the explorer says 9 because 1 ark was sent to themselves 🤔 . At least for the `all` tab it would make sense to have, I don't think `sent` and `received` need it as sent contains the full amount anyway and the received amount should be obvious

dated:
Like this? :hmhm:

image

ItsANameToo:
image
🤔 can you hide the icon if the multipayment doesn't contain transfers to the sender? Wallet address to check is `DTRdbaUW3RQQSL5By4G43JVaeHiqfVp9oh`

dated:
Yea, that was my Intention. I'll fix the condition in a bit.

ItsANameToo (approved):

ghost:

ghost:
Your pull request has been merged and marked as tier 2. _It will earn you $100 USD._

resources/views/components/tables/rows/desktop/address.blade.php
```diff
@@ -1,7 +1,7 @@
 <x-general.identity :model="$model" :without-truncate="$withoutTruncate ?? false">
     @if ($model->username())
         <x-slot name="suffix">
-            <span class="hidden ml-1 lg:flex">
+            <span class="hidden ml-1 font-semibold lg:flex text-theme-secondary-500">
```
Highjhacker, line 4: 🤔 Will have a look, just picked the color directly from the XD

Highjhacker: Yeah should be `secondary-500` (#A5ADB9) there too, will update

PR #783, opened 2019-11-29
feat: use Inter as font

ItsANameToo:
Replaces Proxima Nova with Inter

- [x] figure out why base size is stuck on 16px even though it's defined as 15px


ItsANameToo:
back to wip; need to check the influence of `text-base` on the overall sizes in the explorer 🤔

faustbrian (approved):

PR #789, opened 2019-12-06
feat: add check for businesses / bridgechains being enabled

### LINKED ISSUE #788: feat: optional bridgechain / business pages
These pages should only be loaded when they are available on the blockchain the explorer is running on.

ItsANameToo:
## Summary

Adds a check to disable the bridgechains / businesses pages if they are not enabled on the network. Resolves #788 

Draft because it needs tests and a small refactor

## Checklist

- [x] Tests _(if necessary)_
- [ ] Ready to be merged


dated:
You can probably reduce the `hasBusinessesEnabled` and `hasBridgechainsEnabled` variables down to a single `hasMagistrateEnabled`, since core either adds both resources to the API or none - they aren't added individually.

https://github.com/ArkEcosystem/core/blob/33256dbe9c53334611e8607d40e975bf8184a5d2/packages/core-api/src/handlers/index.ts#L25-L28

ItsANameToo:
ah, thanks @dated

PR #791, opened 2019-12-09
fix: check that sender is requested address

dated:
## Summary

Additionally checks that the sender of the transaction is equal to the requested address when calculating the partial amount sent to sender.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


ghost:
**Thanks for submitting this pull request!** A maintainer will review this in the next few days and explicitly select labels so you know what's going on.

ItsANameToo (approved):

ghost:

ghost:
Your pull request has been merged. Thank you for your contribution to the ARK Ecosystem.

app/Models/Block.php
```diff
@@ -95,4 +96,12 @@ public function getConnectionName()
     {
         return 'explorer';
     }
+
+    /**
+     * Used to force a query with no results.
+     */
+    public function scopeEmpty(Builder $query): Builder
+    {
+        return $query->whereNull('id');
```
ItsANameToo: is this the preferred way to return an empty result? 🤔

app/Services/Search/BlockSearch.php
```diff
@@ -24,11 +27,16 @@ public function search(array $parameters): Builder
...
         if (! is_null($term)) {
-            $query = $query->whereLower('id', $term);
-
-            $numericTerm = filter_var($term, FILTER_VALIDATE_FLOAT, FILTER_FLAG_ALLOW_THOUSAND);
+            if ($this->couldBeBlockID($term)) {
+                $query = $query->whereLower('id', $term);
+            } else {
+                // Forces empty results when it has a term but not possible
+                // block ID
+                $query->empty();
```
ItsANameToo, line 35: why do we need to force empty here?

app/Models/Block.php
ItsANameToo: hmm I see now that you use these in `else` statements, but wouldn't it suffice to not create a query if the `if` does not hit? e.g. here https://github.com/ArkEcosystem/explorer.ark.io/pull/791/files#diff-b59c182bce7d848ab8acb5b96e78d56938d6ac1bf6128802f73eee834b288199R35

alfonsobries: @ItsANameToo the problem is that if the empty scope is not applied it will return all the blocks, so if my search query is, let's say, a random string like `"asdfg"` it incorrectly will show all the blocks since the resulted query will be `select * from "blocks"`. What I'm doing is basically replacing the `select * from "blocks" where lower(id) = "asdfg"` that takes approx `4s` (and we know will not return any result) for  a simple `select * from "blocks"  where id = null` that just takes a few milliseconds and also returns empty results

The problem is that I cannot just return an empty array, I still need a query builder instance since more conditions may apply, in other words, the query is just ensuring that we will not find any blocks by the given id by doing the simplest as fastest query I found. Since the following conditions (that may or not may apply) are `orWhere` still will be possible to find results, for example, if the query is a valid wallet username.

app/Services/Search/BlockSearch.php
alfonsobries: See comment above

app/Models/Block.php
alfonsobries: @ItsANameToo regarding whether that's the preferred way I consider different solutions and I didn't find a better one. I'll revisit this PR in a moment to see if I come up with a cleaner alternative.

alfonsobries:
@ItsANameToo didn't change the solution of empty scopes that are still open for discussion but refactored the empty scopes for something much cleaner. See: https://github.com/ArkEcosystem/explorer.ark.io/pull/791/commits/15eb5e3409a0483b4bd63909b12da1bfd8adc596

ItsANameToo:
added a clickup task to reconsider those empty scopes https://app.clickup.com/t/m97mgv . Still feels like there's a better approach but at this time I cannot think of one.

Issue #776, opened 2019-11-21
feat: pagination for standy / resigned delegates


ItsANameToo:
It would make sense to have pagination for the standby / resigned delegate tables similar to how we paginate the other tables.

PR #792, opened 2019-12-10
feat: delegate pagination

ItsANameToo:
## Summary

Add pagination to the delegate monitor tables so we can see more than just the active / standby / first page of resigned delegates

Left todo:

- [x] update resigned delegate tab to work with the new API response structure
- [x] add mobile version of the table
- [x] improve the table checks for determining between regular / resigned delegates
- [x] think about making these pages fit 51 (depending on network) items per page to match the delegate monitor, or keep it at 25 and show e.g. 100-125 instead of 102-127 when clicking on `show more` on standby delegates
- [x] add `show more` button to standby / resigned delegates when applicable
- [x] add vuex store for the delegate table sorting options
- [x] unit & e2e tests

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/__tests__/e2e/specs/delegate-monitor_spec.js
+++ b/__tests__/e2e/specs/delegate-monitor_spec.js
@@ -236,5 +236,15 @@ describe("Delegate Monitor", () => {
           cy.url().should("include", "wallets/");
         });
     });
+
+    it("should be possible to click the 'show more' button for active delegates", () => {
+      cy.get("h1")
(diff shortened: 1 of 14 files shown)


ItsANameToo:
Leaving vue component testing as a separate task as none of the tables are currently tested yet 😢

ItsANameToo:
Resolves #776

Issue #771, opened 2019-10-28
Docs: backwards compatibility


ItsANameToo:
Check backwards compatibility of the new explorer and core 2.5, documenting any breaking changes.

PR #797, opened 2019-12-13
feat: add changelog

ItsANameToo:
adds a changelog, making it easier to compare differences between versions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,38 +9,38 @@ and this project follows the major and minor version releases of [Core](https://
 
 ### Added
 
-- feat: add changelog (#797)
-- feat: delegate pagination (#792)
-- feat: show registered businesses (#778)
-- feat: bignumber arithmetic (#777)
-- feat: add tab for resigned delegates (#772)
-- feat: support for core 2.6 transactions (#760)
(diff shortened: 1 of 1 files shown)


ItsANameToo:
Resolves #771

alexbarnsley:
should be done @samharperpittam

resources/views/app/block.blade.php
```diff
@@ -22,7 +22,7 @@
                 <x-ark-container>
                     <div id="transaction-list" class="w-full">
                         <div class="flex relative justify-between items-end mb-3">
-                            <h2>@lang('pages.block.transactions')</h2>
+                            <h2 class="text-2xl">@lang('pages.block.transactions')</h2>
```
ItsANameToo: use proper header classes, either change it to `h3` or whatever is needed, or use `header-3` if the tag should be different from the styling
image

ItsANameToo: @alexbarnsley above applies to a few files

alexbarnsley: @ItsANameToo it's forced font size because they're 24px even on mobile devices. I didn't check the new ones Sam recommended though. I will check the things Sam asked to be changed too. Below is from the XD for the Block transactions headings

Largest device:
image

Smallest device:
image

PR #798, opened 2019-12-13
chore: update dependencies

### LINKED ISSUE #795: chore: update dependencies
There are a couple dependencies that need updating, including some major version bumps that need proper testing.

ItsANameToo:
Updates the dependencies to their latest versions. This also includes core-js v3 and vue-cli (+ plugins) v4 (also checked https://cli.vuejs.org/migrating-from-v3/, but looks like we're not affected by most changes).

Would be great if you could go over it too @faustbrian 

Resolves #795


ItsANameToo:
will have to check why cypress failed 🤔 ran fine locally

ItsANameToo:
https://app.clickup.com/t/m36q0j

PR #804, opened 2020-01-02
feat: advanced search

Lemii:
## Summary

This PR adds an advanced search page to the explorer, as described in the following Tier-0 Bounty: [[explorer][750 USD] Advanced Search Page](https://github.com/ArkEcosystem/tier-0-program/issues/8)

Apart from the implementation of the new functionality, there was also some minor refactoring done. For example, the `IApiDelegateVotersWrapper` has been refactored to `IApiWalletsWrapper`, as it can accomodate both objects. Also, the wallet search service has been adjusted to make more similar to the other already existing services.

## Checklist

- [x] Tests _(if necessary)_
- [ ] Ready to be merged


ghost:
**Thanks for submitting this pull request!** A maintainer will review this in the next few days and explicitly select labels so you know what's going on.

ItsANameToo:
@Lemii can you fix the conflicts

src/utils/input-processor.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const inputProcessor = (inputName: string, inputValue: any): string | number => {
+  let value = inputValue;
+
+  const arktoshiValues = ["amount", "totalAmount", "fee", "totalFee", "reward", "balance"];
+  const publicKeyValues = ["vote", "generatorPublicKey"];
+
+  if (arktoshiValues.find(name => inputName.includes(name))) {
+    const valAsNum = Number(value);
+    value = valAsNum > 0 ? Number(value) * 1e8 : 0;
```
ItsANameToo: you can already return the value here instead of it checking the other `if` statements too when it has already found a match

src/utils/input-processor.ts
```diff
@@ -0,0 +1,39 @@
...
+  return lookup ? lookup.publicKey : null;
+};
+
+export const inputProcessor = (inputName: string, inputValue: any): string | number => {
+  let value = inputValue;
+
+  const arktoshiValues = ["amount", "totalAmount", "fee", "totalFee", "reward", "balance"];
+  const publicKeyValues = ["vote", "generatorPublicKey"];
+
+  if (arktoshiValues.find(name => inputName.includes(name))) {
```
ItsANameToo, line 41: why an `includes` instead of an `===`? what are the names of the inputs? 🤔

ItsANameToo (changes requested):

ItsANameToo:
some other things I ran into when I tried to use the search:

- transactions/blocks/wallets; console gives a warning about wallets being `""`, this probably means they are rendered before they are actually loaded properly
- transaction: when I search for `31c17d3a76692a111156e59d605caff78bd359eaca50b7abe24fa4be86e7e26f` it finds one transaction (expected), but if I changed `amount from` to `1`, it suddenly no longer finds anything
- can you set a default value in the transaction type dropdown (just transfer by default)
- wallets; when you search for something that has no results, the table headings jump around for a second
- wallets; when you use a delegate that does not exist with `voting for` (e.g. `itsanametoo1234`), it returns all wallets instead of none. Same happens in the block search when filling in a non-existent delegate in the `generated by` field.
- wallets; when you start typing in the `balance from` field and enter a `0`, it throws an error in the console (e.g. if you want to search for 0.001 as balance). Also happens with other number fields like `total fee`.
- the sort settings for the wallet results are shared with the voters table and top-wallets tables and can result in errors in the console (e.g. set top-wallets to sort by `rank`) and then search for wallets and it will throw an error about `sortFn` being undefined as that column does not exist there.

Other than that, it works great overall. Only some minor issues that need to be looked at

src/utils/input-processor.ts
ItsANameToo: also if `inputName` would match one of the options, it would be possible to use `arktoshiValues.includes(inputName)` instead

Lemii: The reason for this is because the `inputName` that is passed to the function does not always exactly match the entries in that arktoshi/publickey value arrays. This is the case for fields that have a 'from' and 'to' child. For example, when processing the 'amount from' value, the inputname would be `amount-from`.

The component that utilizes this input processor could potentially be refactored to accomodate for this suggestion, if needed.

Lemii:
> some other things I ran into when I tried to use the search:
> 
> * transactions/blocks/wallets; console gives a warning about wallets being `""`, this probably means they are rendered before they are actually loaded properly

Still need to look into this one. It appears that the router falls back to a 2.0 redirect for some reason and therefore expects an address parameter..

```
    {
      path: "/address/:address",
      redirect: to => ({
        name: "wallet",
        params: { address: to.params.address },
      }),
      meta: { title: (route: Route) => getTitle("Wallet") },
    },
```

> * transaction: when I search for `31c17d3a76692a111156e59d605caff78bd359eaca50b7abe24fa4be86e7e26f` it finds one transaction (expected), but if I changed `amount from` to `1`, it suddenly no longer finds anything

This ID is of a multipayment transaction. The actual amount of that transaction itself is 0. The API does not query the sum of the multipayments that are inside of it.

> * can you set a default value in the transaction type dropdown (just transfer by default)
> * wallets; when you search for something that has no results, the table headings jump around for a second
> * wallets; when you use a delegate that does not exist with `voting for` (e.g. `itsanametoo1234`), it returns all wallets instead of none. Same happens in the block search when filling in a non-existent delegate in the `generated by` field.
> * wallets; when you start typing in the `balance from` field and enter a `0`, it throws an error in the console (e.g. if you want to search for 0.001 as balance). Also happens with other number fields like `total fee`.
> * the sort settings for the wallet results are shared with the voters table and top-wallets tables and can result in errors in the console (e.g. set top-wallets to sort by `rank`) and then search for wallets and it will throw an error about `sortFn` being undefined as that column does not exist there.
> 
> Other than that, it works great overall. Only some minor issues that need to be looked at

These issues have now been resolved.

src/pages/AdvancedSearch.vue
```diff
@@ -0,0 +1,257 @@
...
+
+      <div class="hidden sm:block" v-if="selectedType === 'wallet'">
+        <TableWalletsSearchDesktop
+          :wallets="data"
+          :total="supply"
+          :sort-query="sortParams"
+          @on-sort-change="onSortChange"
+        />
+      </div>
+      <div class="sm:hidden">
```
ItsANameToo:
```suggestion
      <div class="sm:hidden" v-if="selectedType === 'wallet'">
```

ItsANameToo:
I looked into the router issue. It's not trying to use the fallback, but this one:

```
{
      path: "/wallets/:address",
      name: "wallet",
      component: WalletComponent,
      meta: { title: (route: Route) => getTitle("Wallet " + route.params.address) },
    },
```

which is used by components such as `LinkWallet`. The issue you're having is that the `AdvancedSearch` component lists all options (transactions, blocks, wallets), but is missing a `v-if` for the mobile wallet search. So on every search you do, it will try to create that in the DOM (but hidden)and since you might be searching for transactions, it can't link to a proper wallet and ends up giving `""` to the component. I've added a comment that should fix the issue

Only thing remaining then is that if you search for a wallet, the `voting for` can be empty and it will still pass `""` to the router, so it will need a check there if the wallet is actually voting so it doesn't unnecessarily links to routes

ItsANameToo:
Regarding the multipayment amount issue, good catch I hadn't linked the two :trollface: Will tell the core guys to look into that

src/utils/input-processor.ts
ItsANameToo: I see, you can leave it as is in that case 🤔

src/pages/AdvancedSearch.vue
```diff
@@ -187,11 +187,20 @@ export default class AdvancedSearchPage extends Vue {
...
     const { name, value } = input;
     let processedName = name;
     let processedVal = inputProcessor(name, value);
 
+    // Fallback
+    if (!processedVal && ["vote", "generatorPublicKey"].includes(name)) {
+      try {
+        processedVal = (await WalletService.find(value)).publicKey;
+      } catch (error) {
+        processedVal = value;
```
faustbrian: This should be `processedVal = undefined;` or you end up using the value that is not a public key to perform the search.

Lemii: This was done intentionally and on request by the maintainer to let the search query show 'no results' if no public key is found for the entered username. If the value would be `undefined`, then the search query would simply return all results.

src/store/modules/ui.ts
```diff
@@ -257,6 +271,11 @@ const getters: GetterTree<IUiState, {}> = {
     return params ? JSON.parse(params) : { field: "originalIndex", type: "asc" };
   },
 
+  walletSearchSortParams(state) {
+    const params = state.walletSearchSortParams || localStorage.getItem("walletSearchSortParams");
+    return params ? JSON.parse(params) : { field: "originalIndex", type: "asc" };
```
ItsANameToo: missed it last time, but `originalIndex` does not exist on the wallet search table (so it will thrown an error in the console if you don't manually sort a column first). It would be better to use `balance` as that's the default by which the API returns the wallets too

ItsANameToo: `balance` should then be `desc` too

src/pages/AdvancedSearch.vue
```diff
@@ -187,11 +187,20 @@ export default class AdvancedSearchPage extends Vue {
...
     }
   }
-  private onFormChange(input: { name: string; value: string }) {
+  private async onFormChange(input: { name: string; value: string }) {
     const { name, value } = input;
     let processedName = name;
     let processedVal = inputProcessor(name, value);
 
+    // Fallback
+    if (!processedVal && ["vote", "generatorPublicKey"].includes(name)) {
```
ItsANameToo: One thing with the fallback in this location, is that `inputProcessor` has a `return publicKey ? publicKey : value;` so it will always return a value and never trigger this code? It would be better to have it inside the `lookupPublicKey` method and use the fallback in case there was no successful lookup.

ItsANameToo (approved):

ghost:

ghost:
Your pull request has been merged but was not assigned a bounty tier. A maintainer will assign a bounty tier to this pull request in the next few days.

ItsANameToo:
still getting a 404 with this @alfonsobries , happened when the active tab is open and the round changes
image

However, didn't seem to run into the slot issue

alfonsobries:
@ItsANameToo ok so I am leaving this PR as the solution for https://app.clickup.com/t/m1datc only

app/Services/Cache/MonitorCache.php
```diff
@@ -21,7 +21,7 @@ public function setBlockCount(Closure $callback): string
         return $this->remember('block_count', Network::blockTime(), $callback);
     }
 
-    public function setNextDelegate(Closure $callback): WalletViewModel
+    public function setNextDelegate(Closure $callback): ? WalletViewModel
```
faustbrian, line 24:
```suggestion
    public function setNextDelegate(Closure $callback): ?WalletViewModel
```

PR #832, opened 2020-01-22
fix: default to address link

### LINKED ISSUE #823: refactor: default to showing the linked address in LinkWallet
If a bridgechain adds their own transaction type, it will not match with any of ours and as a result `LinkWallet` will generate an empty result. Instead, it should simply link to the wallet if it does not match any of the types to provide a minimal usability for unknown transaction types.

ItsANameToo:
When an unknown type / typegroup combination was given, `LinkWallet` would display nothing at all. This refactors the `LinkWallet` component to default to showing the address / public key that was given when nothing matches. Also a small code refactor so we don't end up with duplicate code in `LinkWallet` to accommodate for the default case.

Resolves #823


ItsANameToo:
Needs a testcase to show correct usage for an unknown tx type

ItsANameToo:
https://app.clickup.com/t/mb5p9v

ItsANameToo:
@alfonsobries your before/after pics are the wrong way around

PR #833, opened 2020-01-23
chore: replace tslint with eslint

### LINKED ISSUE #829: chore: setup eslint
tslint is no longer maintained and should be replaced with eslint with a typescript config + have it lint on commit

Highjhacker:
## Summary

Replace `tslint` for `eslint`

Resolve #829 

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

- Not sure about the rules, do we want more than these ?
- Tests were excluded from tslint.json previously, but I included them this time, can remove if not necessary.
- Put it in --no-fix in package.json for now since wasn't sure about the rules we want to apply.


ItsANameToo:
you didn't apply the merge properly @Highjhacker

.eslintignore
```diff
@@ -2,4 +2,4 @@
 /config/
 /dist/
 /*.js
-/__tests__/unit/coverage/
+/node_modules/**
```
ItsANameToo, line 5: please keep the coverage line too in case someone ran it locally and the coverage folder was generated

package.json
```diff
@@ -7,7 +7,7 @@
     "serve": "vue-cli-service serve",
     "build": "vue-cli-service build",
     "test": "yarn test:unit && yarn test:e2e",
-    "lint": "vue-cli-service lint && prettier --write \"./*.{ts,js,json,md,vue}\" \"./src/**/*.{ts,js,json,md,vue}\" \"./__tests__/**/*.{ts,js,json,md,vue}\"",
+    "lint": "vue-cli-service lint --no-fix \"./*.{ts,js,json,md,vue}\" \"./src/**/*.{ts,js,json,md,vue}\" \"./__tests__/**/*.{ts,js,json,md,vue}\"",
```
ItsANameToo: re-add `prettier` to the `lint` command as it's now unused. Also make sure eslint fixes the errors that it can

ItsANameToo (changes requested):
can you run `yarn lint` and see what errors it runs into? then check for the errors whether they are relevant or if the rule should be disabled

Highjhacker:
Better but still linting issues, such as : 

```js
importMultiPaymentTransactions from '@/components/tables/MultiPaymentTransactions.vue';
```
```js
mport CryptoCompareService from '@/services/crypto-compare';
```
```js
export default classstBlocks extends Vue {
```

Also with the spread operator `...` where sometimes it changes it to `..`

ItsANameToo:
when I run this, it still shows a few warnings / errors 🤔

.prettierrc.json
```diff
@@ -1,7 +1,9 @@
 {
-  "printWidth": 120,
+  "bracketSpacing": true,
+  "printWidth": 80,
```
ItsANameToo: better to set this to 120 again, 80 is quite small

.eslintrc.js
```diff
@@ -1,25 +1,29 @@
...
+  ],
   rules: {
-    // allow paren-less arrow functions
-    "arrow-parens": 0,
-    // allow async-await
-    "generator-star-spacing": 0,
-    // allow debugger during development
-    "no-debugger": process.env.NODE_ENV === "production" ? 2 : 0,
-    "vue/component-name-in-template-casing": ["warn", "PascalCase", {}],
+    "no-console": process.env.NODE_ENV === "production" ? "error" : "off",
```
ItsANameToo: I rather not have those production checks, as that will result in issues when people don't lint it with the right `NODE_ENV`

src/App.vue
```diff
@@ -2,7 +2,7 @@
   <main
     :class="[
       nightMode ? 'theme-dark' : 'theme-light',
-      'bg-theme-page-background text-theme-text-content min-h-screen font-sans xl:pt-8',
+      'bg-theme-page-background text-theme-text-content min-h-screen font-sans xl:pt-8'
```
ItsANameToo: can you check about a rule to not remove these commas?

ItsANameToo (changes requested):

ItsANameToo:
Can you also add a lint step to the github workflow so we catch any linting issues in case people skip the step? [desktop wallet has one](https://github.com/ArkEcosystem/desktop-wallet/blob/develop/.github/workflows/test.yml) and OC too which also includes committing any fixes

ItsANameToo (approved):

PR #835, opened 2020-01-24
fix: cypress verify failure

ItsANameToo:
~Applies a workaround to hopefully have the cypress tests run again, based on https://github.com/cypress-io/cypress/issues/6184#issuecomment-577308973~

They just released their new version, so updating to that will fix the issue

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -21,11 +21,6 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
-      - name: Apply Cypress Workaround
-        run: |
-         Xvfb :1 -screen 0 1280x800x24 -nolisten unix &
-         export DISPLAY=:1
-
       - name: Install
         run: yarn install --frozen-lockfile
 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -23,7 +23,7 @@ jobs:
 
       - name: Apply Cypress Workaround
         run: |
-         Xvfb :1 -screen 0 1280x800x24 &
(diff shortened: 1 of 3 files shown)


leMaur:
@alfonsobries It sounds good, even compared to Figma! 👍🏽

alexbarnsley:
@alfonsobries I'm going to mark as draft until the the UI dep is updated. I'd have updated it but I have a PR that also updates that dependency so figured I'd just wait

alfonsobries:
Just updated the ui dependency should be ready now

resources/views/components/general/search/search.blade.php
```diff
@@ -8,7 +8,7 @@
     @search-type-changed.window="searchType = $event.detail"
     class="w-full"
 >
-    <div class="flex flex-col bg-white rounded-xl dark:bg-theme-secondary-900">
+    <div class="flex flex-col bg-white rounded-lg md:shadow-lg-smooth dark:bg-theme-secondary-900">
```
ItsANameToo: shouldn't this be rounded-xl? afaik we don't use rounded-lg much/at all

alfonsobries: @ItsANameToo the design uses less roundiness: 

### Current implementation (with xl)
image

### With `lg` roundiness

image


### Design
image

alfonsobries: @ItsANameToo so according to Oleg lg is correct

ItsANameToo: @alfonsobries according to oleg it should be 12px, so `xl` :trollface: I have updated it on your PR

PR #863, opened 2020-02-10
chore: remove lint-staged in favour of github actions

ItsANameToo:


alexbarnsley:
@ItsANameToo @samharperpittam could you have a look at this please when you get a chance. I've changed it so there's no bottom padding/spacing on devices < 768. This might nee to be approved as it strays from the design

ItsANameToo:
@alexbarnsley you don't really notice that lack of padding, but results in the button not being pushed up as soon as before :oleglala:

leMaur:
@alexbarnsley It seems good to me! 

image

PR #868, opened 2020-02-12
chore: changelog date

ItsANameToo:


alfonsobries:
@alexbarnsley good catch, fixed ✅

PR #876, opened 2020-02-17
feat: show wallet magistrate details

dated:
## Summary

Adds magistrate details (business / bridgechains) to the wallet page.

details

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


ItsANameToo:
what's the status on this @dated

dated:
> what's the status on this @dated

ready for review @ItsANameToo

ItsANameToo:
added `on hold` as there are some considerations I have to make first

ItsANameToo:
sorry @dated , but due to an upcoming project from us this will see some changes. We'll be in touch

app/DTO/Slot.php
```diff
@@ -127,15 +127,7 @@ public function status(): string
...
-        if ($this->isNext()) {
-            return true;
-        }
-
-        if ($this->isPending()) {
-            return true;
-        }
-
-        return false;
+        return $this->isNext() || $this->isPending();
```
ItsANameToo: keep this separate lines for proper coverage

tests/Feature/Http/Livewire/DelegateMonitorTest.php
```diff
@@ -69,12 +69,53 @@ function createRoundWithDelegates(): void
...
+    });
+
+    $height = 1;
+    $wallets->take(5)->each(function ($wallet) use (&$height) {
+        Block::factory()->create([
+            'height'               => $height,
+            'generator_public_key' => $wallet->public_key,
+        ]);
+
+        $height++;
```
ItsANameToo: so you set a block here for 5 wallets at 5 different heights meaning that 5 delegates will have forged

tests/Feature/Http/Livewire/DelegateMonitorTest.php
```diff
@@ -69,12 +69,53 @@ function createRoundWithDelegates(): void
...
+
+    $instance = Livewire::test(DelegateMonitor::class)
+        ->call('pollDelegates')
+        ->instance();
+
+    $delegates = collect($delegateProperty->getValue($instance));
+
+    expect($delegates)->toHaveCount(51);
+
+    $forgedDelegates = $delegates->splice(0, 5);
```
ItsANameToo, line 162: so this should end up being correct as those have a block and it being forged

tests/Feature/Http/Livewire/DelegateMonitorTest.php
```diff
@@ -69,12 +69,53 @@ function createRoundWithDelegates(): void
...
+    $missedDelegates = $delegates->splice(0, 5);
```
ItsANameToo, line 164: this doesn't hold up unless we're at height 10 or so 🤔 I don't see that being set anywhere. So for the test we will be at height 5 in terms of blocks (5 forged) and rest will not have forged yet so they are pending until the calculations determined they missed their slot (I don't recall the exact calculation for this in explorer, you'll have to search)

tests/Feature/Http/Livewire/DelegateMonitorTest.php
```diff
@@ -69,12 +69,53 @@ function createRoundWithDelegates(): void
...
+
+    // dd($delegates->pluck('status'));
+    // dd($delegates, $forgedDelegates, $missedDelegates);
+
+    $forgedDelegates->each(fn ($delegate) => expect($delegate->hasForged())->toBeTrue());
+    $missedDelegates->each(fn ($delegate) => expect($delegate->justMissed())->toBeTrue());
+    $delegates->each(fn ($delegate) => expect($delegate->isWaiting())->toBeTrue());
```
ItsANameToo: not sure what it currently fails on because I can no longer see the logs (too old on this run), can you explain what the issue was with the current test @alexbarnsley ?

tests/Feature/Http/Livewire/DelegateMonitorTest.php
ItsANameToo: what the test should focus on is having forged, missed, missed and checking if those statuses get updated properly, and then have another block being forged as that was where the UI would update properly. So e.g.:

- forge blocks 1-5, 5 delegates will have `hasForged = true`
- two delegates will now miss a block , so there is no block 6 yet but 2 delegates have missed their slot. You'll need to update the internals to reflect this (either in terms of passing time or whatever we use to determine when a slot is missed)
- now assert if the delegate statuses for those 2 delegates are updated properly (if this does not appear to be an issue, then it will be a tricky thing to resolve as it could indicate a pure UI issue where livewire fails to update properly but I doubt that)
- forge a new block (height 6, will be by delegate 8) and check if delegate statuses are still updated properly

app/DTO/Slot.php
alexbarnsley: that's a good call

PR #880, opened 2020-02-25
chore: whitelist status-became-active class

ItsANameToo:


ItsANameToo:
tests fail due to the needed migration to work with the honeypot component but we don't use that on explorer. Waiting on a better resolution in the UI lib before merging this as there's no need to run useless migrations 🤔

PR #941, opened 2020-06-23
fix: properly convert uint8arrays to strings

ItsANameToo:
the buffer change to uint8arrays broke the multisig address generation as uint8arrays can't be changed to hex strings directly


src/mixins/crypto.ts
```diff
@@ -17,12 +17,16 @@ const privateKeyFromPassphrase = (passphrase: string): { publicKey: string; priv
...
 
   return {
-    publicKey: secp256k1.publicKeyCreate(privateKey, true).toString("hex"),
+    publicKey: stringToHex(secp256k1.publicKeyCreate(privateKey, true)),
     // @ts-ignore
     privateKey: privateKey.toString("hex"),
   };
 };
 
+const stringToHex = (text: string): string => {
```
faustbrian:
```suggestion
const stringToHex = (text: string): string => Buffer.from(text).toString("hex");
```

PR #975, opened 2020-10-08
refactor: hydrate wallets to respect the data structure contract

faustbrian:
Instead of adjusting the whole explorer to the new data structures, it makes more sense to use a DTO-ish approach by normalizing the data before it leaves the service. This will keep the explorer functional without any major changes because we can return wallet objects that match the previous format.


ItsANameToo (approved):

ItsANameToo:
tests are rekt

